PR #213, opened 2014-11-28
Add flag disable_sync_compositor to conditional disable sync compositor.

halton:


ds-hwang:
I don't understand why disable_sync_compositor patch is needed. I guess we can control whether enabling sync_compositor by one line patch, instead of this huge patch.
In addition, relevant code is hugely changed already, so this patch became burden to rebase.
could you revert it and commit simple patch?

btw, upstream android goes to the direction not-using sync compositor. why xwalk needs to use sync compositor?

halton:
@ds-hwang current crosswalk-mini branch based on Crosswalk 10.39.x.x, which is 2 milestone late than Crosswalk master, and 3~4 late than chromium upstream. Could you show me how easy one line patch based on current branch?

And thanks for the update of upstream android. Does chromium upstream has removed the sync compositor? If so, I think we can simply remove this commit on next rebase.

ds-hwang:
@halton upstream don't remove sync compositor but don't use it by default now.

I think If SyncCompositorFactory is not initialized, SyncCompositor is not used. so all we need to do is to comment out SyncCompositorFactory initialization or similar branching out.

halton:
@ds-hwang Seems there is mis-understanding of CrosswalkMini branch. We're trying to reduce binary size, disable feature is one of the way.

This PR it will reduce ~20k(not so big). Comment out SyncCompositorFactory initialization won't help out the binary size. The sync compositor code will be built and linked to libxwalkcore.so even it is not used at all. That's why we add code to disable the build of code.

ds-hwang:
@halton I see. thx for explaining. however, I'm not sure it's proper way because it causes maintenance burden.

halton:
yes, maintainence is the issue, that is why we create seperate branch, we are very careful to introduce that to master

PR #218, opened 2014-12-06
Add support for file-picker

joone:
This is part of file-picker patch, which belongs to Chromium content shell.

Bug=XWALK-950


content/browser/webui/content_web_ui_controller_factory.cc
```diff
@@ -14,6 +14,9 @@
 #include "content/public/browser/web_contents.h"
 #include "content/public/browser/web_ui.h"
 #include "content/public/common/url_constants.h"
+#if defined(OS_TIZEN)
+#include "xwalk/runtime/browser/ui/webui/file_picker/file_picker_ui.h"
```
joone: Is okay to add WebUI files to chromium-crosswalk?
I just tried to separate the change from chromium code.

PR #223, opened 2014-12-24
[Temp] Make unix socket Get{Peer|Local}Address return more rational errors.

wuhengzhi:
Rather than return ERR_NOT_IMPLEMENTED it seems sensible to
return ERR_SOCKET_NOT_CONNECTED in the case that the socket
is not connected, but unconditionally return ERR_ADDRESS_INVALID
if connected (because there's no IP addr or port associated
with either end of the socket).

BUG=431412

Review URL: https://codereview.chromium.org/841993002

Cr-Commit-Position: refs/heads/master@{#310564}

### DIFF
--- a/net/socket/unix_domain_client_socket_posix.cc
+++ b/net/socket/unix_domain_client_socket_posix.cc
@@ -9,6 +9,7 @@
 
 #include "base/logging.h"
 #include "base/posix/eintr_wrapper.h"
+#include "net/base/ip_endpoint.h"
 #include "net/base/net_errors.h"
 #include "net/base/net_util.h"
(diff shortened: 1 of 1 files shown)


halton:
lgtm with nit, use [Temp] prefix in git log to indicate rebase owner to remove it in future.

wang16:
LGTM after fix halton's comment, thanks.

wuhengzhi:
I will add it, thanks.

rakuco:
Make sure to roll `DEPS.xwalk`, and possibly backport this to crosswalk-11.

PR #226, opened 2015-01-08
[Temp] Export ui::NativeDisplayDelegateOzone.

rakuco:
This class is used by ozone-wayland, but building Crosswalk with
`-Dcomponent=shared_library` causes linking to fail because this symbol is
not exported by libozone.so even though it is referenced in
libwayland.so.

The best approach upstream is still being discussed, including using a
different class in ozone-wayland. For now, we need it for XWALK-2571, as
building Crosswalk in shared library mode is part of the solution for
it.


rakuco:
@darktears @kalyankondapally

PR #228, opened 2015-01-13
[Tizen] Get visibilityChange events fired by window focus change.

joone:
The IVI homescreen UI has CSS animations, but it still runs and
    consumes more than 35% CPU power when the UI has been hidden.

    This patch allows the window to get visibilityChange events when
    the window loses focus or gets focus.

    BUG=TC-1899


pozdnyakov:
1) Is it upstreamable?
2) The problem Tizen-specific but the code changes is generic, should it be guarded with #if PLATFORM(TIZEN) ?

rakuco:
How does this relate to crosswalk-project/crosswalk#2801?

joone:
@rakuco  These patches are needed for fixing https://bugs.tizen.org/jira/browse/TC-1899

rakuco:
So both patches are needed? Does the Crosswalk side depend on the chromium-crosswalk side being integrated first? You did not state it anywhere.

joone:
@rakuco Yes, the Crosswalk patch depends on this patch so this should be applied first. In addition, there is another patch of ozone-wayland for this.

PR #229, opened 2015-01-26
Allow universal access from file if flag is set and url is file scheme.

huningxin:
BUG=449075
TEST=content_unittests --gtest_filter=NavigationControllerTest.IsInPageNavigation

Review URL: https://codereview.chromium.org/855883002

Cr-Commit-Position: refs/heads/master@{#313051}
(cherry picked from commit 3d426436e10135a99f5e54fe1135767eb55712db)

### DIFF
--- a/content/browser/frame_host/navigation_controller_impl.cc
+++ b/content/browser/frame_host/navigation_controller_impl.cc
@@ -129,7 +129,9 @@ bool AreURLsInPageNavigation(const GURL& existing_url,
                         // for now.
                         existing_url == GURL(url::kAboutBlankURL) ||
                         existing_url.GetOrigin() == new_url.GetOrigin() ||
-                        !prefs.web_security_enabled;
+                        !prefs.web_security_enabled ||
+                        (prefs.allow_universal_access_from_file_urls &&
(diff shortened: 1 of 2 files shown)


huningxin:
Backport from upstream to fix crosswalk bug https://crosswalk-project.org/jira/browse/XWALK-3344.

huningxin:
@darktears @rakuco , please take a look. Thanks!

darktears:
Sure, I see no problem with the backport, can you motivate the need?

huningxin:
Sure. Mobile Chrome Apps and Cordova 4.0 upstream relies on setting "allow-universal-access-from-file" flag via Crosswalk embedding API to do cross-origin in-page-navigation. However Chromium has a issue https://code.google.com/p/chromium/issues/detail?id=449075 that will kill the renderer process (kill Crosswalk Android as it is using single process model) when see cross-origin in-page-navigation even if the "allow-universal-access-from-file" is set. This bug blocks Mobile Chrome Apps and Cordova 4.0 upstream to use Crosswalk 10/11/12. This patch fixes that issue.

PR #233, opened 2015-02-14
[Android] Backport some upstream commits to fix our device tests.

rakuco:
There have been some changes in M41 that causes our device tests to fail on the ICS devices we have. This pull request backports some fixes I made upstream for them so that we get back to a usable state.

We still upload some binaries (like `device_forwarder`) to the device more often than we actually need (see [this Chromium bug I reported](https://code.google.com/p/chromium/issues/detail?id=458658) for more information), but it is nothing too huge.


rakuco:
@darktears

tmpsantos:
RSBM

PR #235, opened 2015-03-03
[Android] Allow constructor of Notifications on Android.

darktears:
We support notifications in Android but upstream doesn't so we
should make sure they are enabled.

We should really consider if the "desktop" notifications are the way
forward for us in Android now that we have ServiceWorker and
push notifications.

https://codereview.chromium.org/920153002

BUG=XWALK-3653


darktears:
@rakuco

PR #236, opened 2015-03-16
[Backport] Put ChannelMojo initialization before starting in-process-render-thread

leonhsl:
ChannelMojo between Browser&Render will be established after
started InProcessRendererThread, this would make some differences
with before(or ChannelMojo disabled case). Actually it would forbid
sending SyncMessage from InProcessRendererThread::Init(), which is enabled before.

BUG=466925
TBR=morrita@chromium.org,rjkroege@chromium.org

Review URL: https://codereview.chromium.org/1003123002

Cr-Commit-Position: refs/heads/master@{#320586}

### DIFF
--- a/content/browser/renderer_host/render_process_host_impl.cc
+++ b/content/browser/renderer_host/render_process_host_impl.cc
@@ -620,11 +620,17 @@ bool RenderProcessHostImpl::Init() {
     // in-process plugins.
     options.message_loop_type = base::MessageLoop::TYPE_DEFAULT;
 #endif
+
+    // As for execution sequence, this callback should have no any dependency
+    // on starting in-process-render-thread.
+    // So put it here to trigger ChannelMojo initialization earlier to enable
+    // in-process-render-thread using ChannelMojo there.
+    OnProcessLaunched();  // Fake a callback that the process is ready.
+
     in_process_renderer_->StartWithOptions(options);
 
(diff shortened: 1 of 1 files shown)


leonhsl:
For https://crosswalk-project.org/jira/browse/XWALK-3650, this PR backports the fix from Upstream.

halton:
LGTM, ping @rakuco

rakuco:
Can you send it targeting the `master` branch instead? I've pushed M42 to master last Friday so we can target it instead.
Also, does this mean we will not need to disable channel-mojo in Crosswalk on Android anymore?

rakuco:
It looks like ChannelMojo was disabled by default in M42 after https://codereview.chromium.org/1002973002.
Doesn't it make more sense then to backport that revision instead and leave your commit for when we move to M43?

leonhsl:
1.
If this backport
"[Backport] Put ChannelMojo initialization before starting in-process-render-thread"
is applied, no need to disable channel-mojo in Crosswalk anymore.

2.
Yes ChannelMojo impl is still ongoing and unstable so I think we'd better backport the revision to disable ChannelMojo by default.  I'll create a new PR soon.
https://codereview.chromium.org/1002973002.
After this backport, also no need to disable channel-mojo in Crosswalk anymore.

PR #239, opened 2015-03-20
[Android] Enable support for notifications on ICS devices.

rakuco:
Upstream recently enabled support for notifications and Push API on
Android >= Jelly Bean (see commit d948e026, "Exceptions for enabling Web
Notifications and the Push API by default").

Our support for notifications in Crosswalk is currently simpler (more
similar to desktop-style notifications) and works on ICS devices as
well, so we can be more lax about which devices we actually support.

Related to: XWALK-3654


rakuco:
@darktears

rakuco:
I'm not entirely sure this is the right way to go though, as this will probably also enable support for Web Notifications for Service Workers on ICS.

darktears:
No because we don't implement the various overloads.

rakuco:
so lgtm?

darktears:
OFC

PR #240, opened 2015-03-21
[Android] Enable copy picture only in contenteditable

axinging:
In Android, copy/paste picture only isn't being good supported.
In chromium for Android, it supports copy/paste text with picture.
In order to align with Android, chromium for Android disabled
copy/paste picture only previously.

But copy/paste picture only and text with picture among chromium
for Android and Android native applications behave the same.
So we enable copy/paste picture only.

BUG=XWALK-3764


axinging:
@halton @lincsoon @kenchris @darktears

ui/base/clipboard/clipboard_android.cc
```diff
@@ -102,14 +102,6 @@ void ClipboardMap::Set(const std::string& format, const std::string& data) {
...
     Java_Clipboard_setText(env, clipboard_manager_.obj(), str.obj());
   } else if (format == kHTMLFormat) {
-    // Android's API for storing HTML content on the clipboard requires a plain-
-    // text representation to be available as well. ScopedClipboardWriter has a
-    // stable order for setting clipboard data, ensuring that plain-text data
-    // is available first. Do not write to the clipboard when only HTML data is
-    // available, because otherwise others apps may not be able to paste it.
-    if (!ContainsKey(map_, kPlainTextFormat))
-      return;
-
```
halton: I think customer complain that this bug appear on chrome on Android as well, not sure it is suitable to fix it on upstream.

If upstream does not want support this, as for how to fix it. Can we add the key `kPlainTextFormat` in crosswalk?

axinging: I have submitted a bug here at chromium: https://code.google.com/p/chromium/issues/detail?id=469565

rakuco:
Since this also affects Chromium upstream and is not a P1 bug, I'd like to wait for them to provide some feedback and only then proceed with changing our version if necessary.

halton:
@axinging please rebase your code.

PR #244, opened 2015-03-30
Add the command line option to load proprietary codec library.

shaochangbin:
Due to licensing issues, the pre-built binary doesn't ship with the
necessary codecs for patented media formats. Therefore, mp4/mp3 files
can't be played back by Crosswalk.

Add a command line option so that mp4/mp3 files can be played by loading
the proprietary library. An example is,
xwalk --proprietary-codec-lib-path=/opt/google/chrome/libffmpegsumo.so

### DIFF
--- a/content/browser/zygote_host/zygote_host_impl_linux.cc
+++ b/content/browser/zygote_host/zygote_host_impl_linux.cc
@@ -137,6 +137,7 @@ void ZygoteHostImpl::Init(const std::string& sandbox_cmd) {
     switches::kLoggingLevel,
     switches::kNoSandbox,
     switches::kPpapiInProcess,
+    switches::kProprietaryCodecLibPath,
     switches::kRegisterPepperPlugins,
     switches::kV,
     switches::kVModule,
--- a/media/base/media_posix.cc
+++ b/media/base/media_posix.cc
@@ -6,9 +6,11 @@
 
 #include <string>
 
+#include "base/command_line.h"
 #include "base/files/file_path.h"
 #include "base/logging.h"
 #include "base/strings/stringize_macros.h"
+#include "content/public/common/content_switches.h"
 #include "media/ffmpeg/ffmpeg_common.h"
 #include "third_party/ffmpeg/ffmpeg_stubs.h"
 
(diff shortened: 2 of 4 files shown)


shaochangbin:
@rakuco @seanlong @halton   PTAL.
The case is similar with below discussion,
https://github.com/nwjs/nw.js/wiki/Using-MP3-&-MP4-(H.264)-using-the--video--&--audio--tags.

content/browser/zygote_host/zygote_host_impl_linux.cc
```diff
@@ -140,6 +140,7 @@ void ZygoteHostImpl::Init(const std::string& sandbox_cmd) {
     switches::kRegisterPepperPlugins,
     switches::kV,
     switches::kVModule,
+    switches::kProprietaryCodecLibPath,
```
halton: nit, order

media/base/media_posix.cc
```diff
@@ -6,12 +6,15 @@
...
 
+#include "base/command_line.h"
 #include "base/files/file_path.h"
 #include "base/logging.h"
 #include "base/strings/stringize_macros.h"
+#include "content/public/common/content_switches.h"
 #include "media/ffmpeg/ffmpeg_common.h"
 #include "third_party/ffmpeg/ffmpeg_stubs.h"
 
+
```
halton: remove this meaningless and wrong change.

halton:
LGTM with the nits comments.

shaochangbin:
Fix the nits. Thanks for your comments @halton

halton:
LGTM

darktears:
Let's make sure it's documented.

shaochangbin:
@darktears , could you please point out the place where I should put the document?
Thanks!

halton:
https://github.com/crosswalk-project/crosswalk-website/wiki/Crosswalk-Command-Line-Options

shaochangbin:
Thanks! @halton 
Updated the wiki page.

rakuco:
I think this needs to be reverted or reworked.

This has broken the `component=shared_library` build:
```
obj/media/base/media.media_posix.o:media_posix.cc:function media::internal::InitializeMediaLibraryInternal(base::FilePath const&): error: undefined reference to 'switches::kProprietaryCodecLibPath'
obj/media/base/media.media_posix.o:media_posix.cc:function media::internal::InitializeMediaLibraryInternal(base::FilePath const&): error: undefined reference to 'switches::kProprietaryCodecLibPath'
obj/media/base/media.media_posix.o:media_posix.cc:function media::internal::InitializeMediaLibraryInternal(base::FilePath const&): error: undefined reference to 'switches::kProprietaryCodecLibPath'
```

Basically you are using a symbol from content/ in media/ which does not work if you build those targets as separate shared libraries, and you cannot make the latter depend on the former because the former depends on the latter.

PR #248, opened 2015-04-14
[Temp][oz-wl] Fix undefined references for NativeViewAccessibility

mrunalk:
Also fix undefined reference for AXPlatformNode::Create


rakuco:
`#if` checks aren't usually line-wrapped as far as I can remember. Other than that, have you pinged upstream about this?

rakuco:
@kalyankondapally fyi, this is probably needed for standalone oz-wl builds too

rakuco:
I've adopted [@kalyankondapally's solution](https://github.com/01org/ozone-wayland/blob/Milestone-Hood/patches/0014-Fix-linking.patch) instead, and pushed together with a rebased chromium-crosswalk tracking 43.0.2357.18.
We still need to find a proper solution for https://codereview.chromium.org/1005293002 upstream though.

PR #249, opened 2015-04-28
[Android] Enable copy picture only in contenteditable

axinging:
In Android, copy/paste picture only isn't being good supported.
In chromium for Android, it supports copy/paste text with picture.
In order to align with Android, chromium for Android disabled
copy/paste picture only previously.

But copy/paste picture only and text with picture among chromium
for Android and Android native applications behave the same.
So we enable copy/paste picture only.

BUG=XWALK-3764

(cherry picked from commit 5dc950dfd897c45bc83f7250760b8be1eeb66ee5)


axinging:
@halton @lincsoon @kenchris @darktears

halton:
@axinging is this PR just simple rebase of #240? If so, please close either one. (Note, recreate a new PR is not good idea, just simple rebase existing one is preferred.)

axinging:
Thanks @halton 
Hi, guys, please  help to review asap:  @halton @lincsoon @kenchris @darktears

halton:
@axinging this PR is broken, please rebase.

axinging:
OK, now let's  go go go! @halton @lincsoon @kenchris @darktears

ui/base/clipboard/clipboard_android.cc
```diff
@@ -102,14 +102,6 @@ void ClipboardMap::Set(const std::string& format, const std::string& data) {
...
     Java_Clipboard_setText(env, clipboard_manager_.obj(), str.obj());
   } else if (format == kHTMLFormat) {
-    // Android's API for storing HTML content on the clipboard requires a plain-
-    // text representation to be available as well. ScopedClipboardWriter has a
-    // stable order for setting clipboard data, ensuring that plain-text data
-    // is available first. Do not write to the clipboard when only HTML data is
-    // available, because otherwise others apps may not be able to paste it.
-    if (!ContainsKey(map_, kPlainTextFormat))
-      return;
-
```
halton: Neither disable or enable HTML P&P unconditionally does not make any sense . Can we expose this key via `XWalkPreferences`, so that developer can control it via this key.

axinging: Please refer to :https://code.google.com/p/chromium/issues/detail?id=469565

Currently, copy/paste picture only is intentionally disabled. Because of copy/paste picture only isn't being good supported in Android system. The detail behavior of current chrome for android are listed below:
1),Copy picture(s) with text from chrome for Android and pastes it into an Android 
apps, it seldom works,picture(s) turns into "OBJ".
2),Copy picture(s) only from chrome for Android and pastes it into an Android apps, it seldom works,picture(s) turns into "OBJ".
3),Copy/paste picture(s) with text in chrome for Android, it works.
4),Copy/paste picture(s) only in chrome for Android, it doesn't work.

It's poor user experience for item 3) and item 4). 
So this fix is intent to  align item 3) and 4).

axinging:
The chromium related commit: https://codereview.chromium.org/25668005

axinging:
@halton, please help to review.
In https://codereview.chromium.org/25668005, it said very clearly that:
"Do not write to the clipboard when only HTML data is available, because otherwise others apps may not be able to paste it."
Indeed in current implementation, copy text+picture from chromium to android native view will not succeed. the picture will turn into OBJ.

halton:
@axinging thanks for update, please rebase your code.

halton:
ping @axinging again to rebase so that we can merge this fix.

axinging:
https://github.com/crosswalk-project/crosswalk/pull/3173

halton:
LGTM, @darktears @rakuco please help to merge.

@axinging the rolling merging commit, which the SHA1 will be generated automatically  after PR merged.

rakuco:
Could you please explain why this does not seem to have been submitted upstream? I see the open bug there, but no related CL.

axinging:
@rakuco  I have a CL for this: https://codereview.chromium.org/1139903002/. Currently there isn't tests for  clipboard android(dcheng). dcheng asked me to enable this tests first. So this was still not got any lgtm on upstream.

rakuco:
OK, then are you going to try enabling the mentioned tests first? This really looks like a proper improvement that would benefit upstream.

axinging:
Hi, @rakuco, how about merge this first as it requested by customer?  Upstream will meaningful,  but it takes time.

rakuco:
+@darktears

The problem with this approach is that it generally means upstreaming never happens because the change has already made it to chromium-crosswalk. We are then forced to carry and maintain it forever. This has already happened with other patches we currently have in the tree.

Since the related bug is not P1 and also affects upstream, I'd really like to see an attempt to fix it there first and only land it here first as a last resort.

axinging:
@halton , how about raise this to P1?

axinging:
@halton , how about raise this to P1? Customer has sent mail to me to  request  this.  Also, halton pinged me for two times to rebase.

Regards,
Xing

halton:
Yes, this is important customer required feature, please prioritize to merge. Add @baleboy in CC loop.

rakuco:
I'm sorry, I might be missing something obvious, but I still don't see how this is different from any other user bug report that is caused by upstream.
Of course I would like to see this fixed, but as I said I don't see this being extreme enough to the point that it makes sense to bypass at least engaging upstream (working on dcheng's suggestion, mentioning this in the android-webview-dev mailing list etc). The priority of a bug should not generally be directly proportional to the number of times people ask about it.
The other points I mentioned about this later becoming a burden we (actually, me and Alexis most of the time) have to carry and maintain and how upstreaming efforts are forgotten once it gets merged here are also very important, and were not refuted.

baleboy:
Agree with @rakuco that this needs to be submitted upstream. There has been plenty of time to do it.

baleboy:
I'd also like to add that our rebases are taking longer and longer for each release, therefore we need to be extremely careful about what extra patches we take in.

PR #250, opened 2015-05-16
M44 Rebase: Add declaration for DictionaryValue

mrunalk:
Without above change I am getting compiler error saying,
../../components/devtools_http_handler/devtools_http_handler.h:124:9: error: no type named 'DictionaryValue' in namespace 'base'


halton:
@mrunalk this seems upstream issue, isn't it?

rakuco:
It does. Mrunal, are you getting this error when building the devtools code in Crosswalk or Chromium? My guess is that the Chromium code works because `base/values.h` ends up being included before this header you are changing and masks the problem.
Can you send this patch upstream first and then just cherry-pick it once it is landed?

mrunalk:
@halton @rakuco Yep, I am going to submit this patch upstream. Just wanted to get a confirmation from you guys. Sorry for not making that clear in the description. Thanks.

PR #251, opened 2015-05-28
Add gyp flag use_minimum_resources and exclude unnecessary resources

lincsoon:
Remove following resources:
1. Resources for media player control bar, this part is not usually
used, thus it could be overriden by web developers.
2. The ui resources are for aura/views, Android doesn't need it.

This patch is the backport from Crosswalk-lite

BUG=https://crosswalk-project.org/jira/browse/XWALK-3928
TEST=Build with/without flag use_minimum_resources and simple test on XWalkCoreShell.apk
SIZE_REDUCED=212k(Release build on ARM)

### DIFF
--- a/content/child/blink_platform_impl.cc
+++ b/content/child/blink_platform_impl.cc
@@ -741,6 +741,7 @@ struct DataResource {
 const DataResource kDataResources[] = {
     {"missingImage", IDR_BROKENIMAGE, ui::SCALE_FACTOR_100P},
     {"missingImage@2x", IDR_BROKENIMAGE, ui::SCALE_FACTOR_200P},
+#if !defined(USE_MINIMUM_RESOURCES)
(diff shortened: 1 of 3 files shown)


content/child/blink_platform_impl.cc
```diff
@@ -741,6 +741,7 @@ struct DataResource {
 const DataResource kDataResources[] = {
     {"missingImage", IDR_BROKENIMAGE, ui::SCALE_FACTOR_100P},
     {"missingImage@2x", IDR_BROKENIMAGE, ui::SCALE_FACTOR_200P},
+    /*
```
halton: comment out is not right way if we consider other platforms like Linux. I'd suggest we use define and use USE_MINIMUM_RESOURCES macro.

halton:
LGTM, ping @darktears

PR #258, opened 2015-07-20
[Lite][Android] Crash found when executes "selection.addRange", caused by icu feature.

heke123:
The line break property for ICU can't be supported without extra libs like ICU4j.
BUG=XWALK-4645

### DIFF
--- a/base/icu_alternatives_on_android/java/src/org/chromium/base/icu/IcuUtils.java
+++ b/base/icu_alternatives_on_android/java/src/org/chromium/base/icu/IcuUtils.java
@@ -311,11 +311,6 @@ public class IcuUtils {
     }
 
     @CalledByNative
-    private static boolean isLineSeparator(int codePoint) {
-        return (Character.getType(codePoint) & Character.LINE_SEPARATOR) != 0;
-    }
-
-    @CalledByNative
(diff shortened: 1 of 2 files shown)


heke123:
@d0ngjun @PeterWangIntel  please help review. thanks.

base/icu_alternatives_on_android/java/src/org/chromium/base/icu/IcuUtils.java
```diff
@@ -310,10 +310,11 @@ private static boolean isPunct(int codePoint) {
         return (Character.getType(codePoint) & flags) != 0;
     }
 
+    /* Currently not be used.
     @CalledByNative
     private static boolean isLineSeparator(int codePoint) {
         return (Character.getType(codePoint) & Character.LINE_SEPARATOR) != 0;
-    }
```
d0ngjun: How about reset the code change in icu_utils.cc and return false here?

heke123:
@d0ngjun,  modification done. please review again, thanks.

d0ngjun:
LGTM

heke123:
@halton  please help review this fix, thanks.

PR #263, opened 2015-08-03
[Backport] Remove NOTIMPLEMENTED() logspam. Provide an implementation…

wuhengzhi:
… that

mirrors what org/chromium/ui/base/Clipboard.java can currently
provide.

Also add testing for ReadAvaliableTypes() and correct bounds error
on Android with UnicodeHTMLTest.

BUG=447167

Review URL: https://codereview.chromium.org/1114213004

Cr-Commit-Position: refs/heads/master@{#339879}

BUG=XWALK-4608

### DIFF
--- a/ui/base/clipboard/clipboard_android.cc
+++ b/ui/base/clipboard/clipboard_android.cc
@@ -315,9 +315,21 @@ void ClipboardAndroid::ReadAvailableTypes(ClipboardType type,
     return;
   }
 
-  NOTIMPLEMENTED();
-
   types->clear();
+
+  // would be nice to ask the ClipboardMap to enumerate the types it supports,
+  // rather than hardcode the list here.
+  if (IsFormatAvailable(Clipboard::GetPlainTextFormatType(), type))
+    types->push_back(base::UTF8ToUTF16(kMimeTypeText));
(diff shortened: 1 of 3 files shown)


wuhengzhi:
@lincsoon @halton @axinging @rakuco Please take a look, should we modify the coding style? Thanks.

halton:
Code LGTM. Since this PR is backport, I think the trybot style failures can be ignored. ping @rakuco

rakuco:
Yeah, some of the style failures do not even come from this patch. However, can't we just wait for this to waterfall to chromium-crosswalk naturally (probably in M46)? It only seems to fix an annoyance, not a severe bug.

wuhengzhi:
@rakuco This patch is needed. The value of ```clipboardData.getData``` will not be got without this patch.
In android, ```console.log("test=", test)``` does not work, but ```console.log("test=" + test)``` works fine.

halton:
This is customer(Youdao note) reported bugs, they're upgrading their apps with latest Lite build. We'd better to merge this PR so that latest Crosswalk lite rebasing have this fix and we can deliver the updated build to YouDao team.

rakuco:
OK. Don't forget to roll `DEPS.xwalk`.

PR #266, opened 2015-08-11
Fix build on Windows when using components build.

darktears:
Due to some ffmpeg changes which now statically build ffmpeg onto
Chromium/Crosswalk the component build was broken on Windows. It turns
out that it couldn't link due to missing symbols in ffmpeg.dll (the very
same we use in sysapp code). It's due to two changes in upstream :
- https://chromium.googlesource.com/chromium/third_party/ffmpeg/+/467a5ef970
which now use ffmpeg.sigs to generate the exported symbols
- https://chromium.googlesource.com/chromium/third_party/ffmpeg/+/59c940d188
which is a following patch after ffmpegsumo removal.

The code we have in chromium-crosswalk which adds the extra symbols
we use from ffmpeg needs to be adapted and the sigs file needs to be
renamed. See original patch :

https://github.com/crosswalk-project/chromium-crosswalk/commit/60bc4428aaa0a10a1522dbbca2796b4352d14747


darktears:
@rakuco

rakuco:
lgtm

rakuco:
does this only affect Windows or do we need this in crosswalk 15 too?

darktears:
Only affect Windows, the def files are only used on Windows...

PR #270, opened 2015-08-25
Fix the conflicts of ResourceExtractor on M45

lincsoon:
The upstream changes from M45 almost make our following commits unnecessary. This patch is to resolve the conflicts and restore previous functionality.

This patch rebased the commits below:
4ba805f1c8d470eb7c32348f2d3c8df28199f7ce
c715ce1b87516f983620c32919a0f99db095bcbe

The commit below can be discared:
31d05593d00b70050a9b4fc6d14f33a5b56fc60e

After this patch get merged, I will send another PR to Crosswalk repository. Then XWALK-4755 can be fixed.

### DIFF
--- a/base/android/java/src/org/chromium/base/ResourceExtractor.java
+++ b/base/android/java/src/org/chromium/base/ResourceExtractor.java
@@ -128,6 +128,11 @@ public class ResourceExtractor {
             try {
                 // Extract all files that don't already exist.
                 for (ResourceEntry entry : sResourcesToExtract) {
+                    // Loading "icudtl.dat" from "assets/"" currently does not work with either
+                    // embedded mode (the file is in raw/res) or shared mode (the app's context is
+                    // used to retrieve the AssetManager, not Crosswalk's). We thus need to put
+                    // those special files in a different directory so that we leverage the fallback
+                    // code in Chromium to load these files from disk.
                     File dir = isAppDataFile(entry.extractedFileName) ? appDataDir : outputDir;
                     File output = new File(dir, entry.extractedFileName);
                     if (output.exists()) {
@@ -322,6 +327,10 @@ public class ResourceExtractor {
 
         try {
             mExtractTask.get();
+            // ResourceExtractor is not needed any more.
+            // Release static objects to avoid leak of Context.
+            sInterceptor = null;
+            sInstance = null;
         } catch (CancellationException e) {
             // Don't leave the files in an inconsistent state.
             deleteFiles();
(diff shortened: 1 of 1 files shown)


lincsoon:
@rakuco Please help to review

halton:
LGTM

rakuco:
Thanks, this version looks so much easier to understand than the previous one.

Is the Crosswalk side already available somewhere for me to take a look? This would probably answer my only question: why is `isAppDataFile()` needed? In the case where no file is intercepted, why should they be installed in a different location (the app dir instead of the output dir)?

lincsoon:
The relevant PR in the Crosswalk side is https://github.com/crosswalk-project/crosswalk/pull/3220

lincsoon:
I don't really know the reason too. It's just that the chromium must search the file `icudtl.dat` from the app dir.

rakuco:
Is that because of some change on our side? Because originally (ie. without our patches) it does not, right?

lincsoon:
Sorry, I have to do more research to answer your question. The works related to ICU data was done by other colleagues that have leaved. But I don't think this file is only for Crosswalk.

halton:
@lincsoon when you talking about ICU, is that related with ICU alternative for the lite build?

lincsoon:
It's not about ICU alternative, but about what does it mean to chromium and how chromium would treat with related file.

rakuco:
I'd also have to investigate it further, but my guess is that if we're behaving differently from Chromium itself it's because something changed upstream and we didn't adapt (like icu being shipped/loaded from a different location), or because we're using icudat differently and really need it from a specific location.

rakuco:
So I did the investigation I talked about. Since it's also related to the PR you sent to Crosswalk, I think it's better to continue discussing in JIRA so everything is centralized in only one place.

base/android/java/src/org/chromium/base/ResourceExtractor.java
```diff
@@ -114,13 +128,19 @@ private void doInBackgroundImpl() {
             try {
                 // Extract all files that don't already exist.
                 for (ResourceEntry entry : sResourcesToExtract) {
-                    File output = new File(outputDir, entry.extractedFileName);
+                    File dir = isAppDataFile(entry.extractedFileName) ? appDataDir : outputDir;
```
rakuco: I think this part of the change deserves an explanation in a comment: as I mentioned in JIRA, we only need to do this (and bring back `isAppDataFile()`) because of what we do with `icudtl.dat` (and the V8 snapshot files if we ever use them on Android): loading them from `assets/` currently does not work with either embedded mode (the file is in `raw/res`) or shared mode (the app's context is used to retrieve the `AssetManager`, not Crosswalk's).
We thus need to put those special files in a different directory so that we leverage the fallback code in Chromium to load these files from disk.

rakuco:
In addition to the comment part I mentioned, please rebase the patch as I've rolled a newer Chromium in the next branch.

rakuco:
Ping

lincsoon:
@rakuco Done!

rakuco:
thanks!

PR #271, opened 2015-08-27
Add gyp flag use_minimum_resources and exclude unnecessary resources

lincsoon:
Remove following resources:
1. Resources for media player control bar, this part is not usually
used, thus it could be overriden by web developers.
2. The ui resources are for aura/views, Android doesn't need it.

This patch is the backport from Crosswalk-lite

BUG=https://crosswalk-project.org/jira/browse/XWALK-3928
TEST=Build with/without flag use_minimum_resources and simple test on XWalkCoreShell.apk
SIZE_REDUCED=44k(Release build on ARM)
(cherry picked from commit eaacec2898e7f7f4764dd9ea320638b410e4ea07)


lincsoon:
This PR fixes https://crosswalk-project.org/jira/browse/XWALK-4756

lincsoon:
@rakuco @halton

halton:
code LGTM, not sure whether we need leave the "conflits" part in log. @rakuco what's your option?

rakuco:
Have you checked if the size measurements are still valid? The reason the conflicts were there is because a lot of files were moved to a different directory/grd file, so if you're only editing the previous one it's possible that the size benefit is smaller, to the point where it isn't worth it keeping this commit around.

lincsoon:
Thanks for your remind! The shinked size has changed from 212k to 44k.

@halton Please evaluate if we should keep thie commit

halton:
I think we need investigate how to get back the ~200k.

lincsoon:
So do you agree to suspend this pull request and skip this feature in the next branch temporarily? Since this is not the core feature, I thinks it's better not to have this issue block current branching.

halton:
Yes, i agree.

PR #273, opened 2015-09-02
Audio focus cherry pick

biodatacomp:
Cherry-pick of:
https://github.com/scheib/chromium/commit/3d020b1e04e93bbd318105276a89dd3aa4af1a27

*Solves the issue with Audio-focus.
Current version of crosswalk-chromium does not have these changes, part of the changes done to MediaPlayerListener were same/similar to what I had done earlier, this would serve as more complete implementation.

Unit tests done:
1) Audio plays fine on Mobile/Leanback/Amazon devices 
2) Takes Focus from Audio already playing in background
3) On pausing / resuming video in our app, doest not give away Audio-focus


rakuco:
This is a fairly long set of patches, and the final diff is huge. Since this is already present in Chromium 45, I think it makes sense to just wait until we roll 45 into Crosswalk master (should be really soon).

PR #276, opened 2015-09-11
Fix errors in debug building

PeterWangIntel:
This patch consists of two parts: devtools-related and net-related.
All of them are harmless for release version.

### DIFF
--- a/net/log/net_log_util.cc
+++ b/net/log/net_log_util.cc
@@ -203,11 +203,13 @@ scoped_ptr<base::DictionaryValue> GetNetConstants() {
   {
     base::DictionaryValue* dict = new base::DictionaryValue();
 
+#if !defined(DISABLE_QUIC_SUPPORT)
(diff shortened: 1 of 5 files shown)


PeterWangIntel:
@d0ngjun

net/log/net_log_util.cc
```diff
@@ -203,11 +203,14 @@ scoped_ptr<base::DictionaryValue> GetNetConstants() {
   {
     base::DictionaryValue* dict = new base::DictionaryValue();
 
+// TODO (Peter Wang): Remove the temp fix for debug building.
```
d0ngjun: Seems the code below is related with QUIC, you can try to use macro DISABLE_QUIC_SUPPORT to control it.

PeterWangIntel: thx for the info, let me try

d0ngjun:
LGTM

PR #277, opened 2015-09-11
next-m46: buildable

jondwillis:
@mrunalk


rakuco:
The annotations change looks fine, thanks for that. note that it will likely end up being squashed into the original commit in the end, but your contribution is very welcome :-)

As for the lint one, it's not clear from the commit message what exactly it is fixing (and since there's no Crosswalk-specific change to that file whatever's happening is possibly an upstream issue.

PR #280, opened 2015-09-15
Change start-up flag kDisableNotifications from jelly bean to ICS to …

chuan9:
…resolve no notifications on android 4.0

CCBUG=XWALK-4669
Related to:XWALK-3645
It's a regression caused by upstream dropping support for Android ICS and our subsequent patches to reenable it. So I back this patch to crosswalk15


chuan9:
@rakuco I back it to crosswalk15 here.

rakuco:
almost there-- you forgot to use `git cherry-pick -x` like we usually do when backporting a patch from a different branch.

chuan9:
Done, Thanks!

rakuco:
thanks.
don't forget to roll `DEPS.xwalk` in both affected Crosswalk branches.

chuan9:
Got it. Thanks!

PR #281, opened 2015-09-15
Revert "Bug fix for minimum resources"

d0ngjun:
This reverts commit ef3e7c6eadb30fccdd7c0881119d659a5aedad86.

For crosswalk-lite-10, a macro named IDR_UASTYLE_HTML_CSS generate in
./gen/blink/public/resources/grit/blink_resources.h,
whether the option use_minimum_resources is enable or disable.
In crosswalk-lite, IDR_UASTYLE_HTML_CSS would not generate if we disable
use_minimum_resources, so we need add it back to avoid the compiling
error.

Simply remove IDR_UASTYLE_HTML_CSS html.css will cause the layout mess
up.


d0ngjun:
@PeterWangIntel @heke123 @halton please help review.

halton:
LGTM, merging.

PR #282, opened 2015-09-25
[Android] Don't depend on android support v13 library

fujunwei:
The commit "[BackgroundSync] Reland of Launch Chrome on Android for One-Shot" in
https://codereview.chromium.org/1140813006/#ps340001 is
On Android devices BackgroundSync needs to launch the browser the next time the device
goes online if a one-shot sync is registered.
The class of the service will be moved to xwalk/app/android/app_support. It will be compiled
in app-side if customer want to try this feature.

BUG=XWALK-5092

### DIFF
--- a/build/java.gypi
+++ b/build/java.gypi
@@ -56,6 +56,7 @@
     'input_jars_paths': [ '<(android_jar)' ],
     'additional_src_dirs': [],
     'javac_includes': [],
+    'filtered_java_files%': '',
     'jar_name': '<(_target_name).jar',
     'jar_dir': '<(PRODUCT_DIR)/lib.java',
     'jar_path': '<(intermediate_dir)/<(jar_name)',
@@ -260,7 +261,7 @@
       'variables': {
         'extra_args': [],
         'extra_inputs': [],
-        'java_sources': ['>!@(find >(java_in_dir)>(java_in_dir_suffix) >(additional_src_dirs) -name "*.java")'],
+        'java_sources': ['>!@(find >(java_in_dir)>(java_in_dir_suffix) >(additional_src_dirs) -name "*.java" >(filtered_java_files))'],
         'conditions': [
           ['enable_errorprone == 1', {
             'extra_inputs': [
--- a/content/content.gyp
+++ b/content/content.gyp
@@ -440,7 +440,6 @@
             '../ui/android/ui_android.gyp:ui_java',
             '../ui/touch_selection/ui_touch_selection.gyp:selection_event_type_java',
             '../ui/touch_selection/ui_touch_selection.gyp:touch_handle_orientation_java',
-            '../third_party/android_tools/android_tools.gyp:android_support_v13_javalib',
             '../third_party/WebKit/public/blink_headers.gyp:blink_headers_java',
             '../ui/mojo/geometry/mojo_bindings.gyp:mojo_geometry_bindings',
             'common_aidl',
@@ -461,6 +460,7 @@
           ],
           'variables': {
             'java_in_dir': '../content/public/android/java',
+            'filtered_java_files': ' ! -name BackgroundSyncLauncherService.java',
             'has_java_resources': 1,
             'R_package': 'org.chromium.content',
             'R_package_relpath': 'org/chromium/content',


rakuco:
I'd appreciate it if @darktears, @r0b5t4 and @lincsoon could take a look at this patch (I'm on holiday next week). Basically, a patch that comments out a bunch of lines always looks suspicious to me:

* https://codereview.chromium.org/1140813006/ says the feature is enabled behind a flag. Can't the flag just be disabled?
* Why not depending on "android support v13" fix the related JIRA ticket?
* Can't the entire commit be reverted instead?
* Most importantly, what's the long-term plan here? We cannot keep commenting those lines out forever, especially as the related code is likely to change a lot.

fujunwei:
Java Code can't be controlled by flag on compiled time.
> https://codereview.chromium.org/1140813006/ says the feature is enabled behind a flag. Can't the flag just be disabled?

fujunwei:
Eclipse and Android studio will depend on android support v4 by default, so the error message will be outputted if following the documentation https://crosswalk-project.org/documentation/embedding_crosswalk.html.
> Why not depending on "android support v13" fix the related JIRA ticket?

fujunwei:
In this case, Only BackgroundSyncLauncherService.java use the API of android support library and it 's independent that there is no other class to use.
So Does it can be hosted  manually in xwalk_core_library/src directory if customer need use this feature?
>  Most importantly, what's the long-term plan here? We cannot keep commenting those lines out forever, especially as the related code is likely to change a lot.

darktears:
The commit message needs a lot more details especially what @rakuco said. Java stuff can be controlled by flag yes as you can pass command line arguments to XWalk on Android and @rakuco was talking about runtime flag, not compile flag, so does the upstream CL.

 - Can't the entire commit be reverted instead?  -> needs answer.

We don't comment code in general : this is bad practice.

fujunwei:
Of course, we can revert entire commit, but the better approach is keep the feature in the Crosswalk,  the depended classes need to be exported so that customer can use the feature if self-compiling the class in app-side. It's important design for similar feature in the future.  
> Can't the entire commit be reverted instead

fujunwei:
The commit of moving the depended class to xwalk/ app/android/app_support directory is  https://github.com/fujunwei/crosswalk/commit/3395073e0a3f9fad34bf6270523add6b82e1a15d

baleboy:
@darktears @rakuco ping

darktears:
I don't think copying the file from Chromium to Crosswalk is a good idea, especially that the feature is experimental, likely to change and we won't keep up-to-date the copy. Can't we just patch the content.gyp to remove the dependency and content/content_jni.gypi? The feature is maybe useful but it's experimental and not finished, there is no need to make any effort to expose it to "customers".

fujunwei:
This PR is remove the dependency from content.gyp.

fujunwei:
ping @darktears @rakuco

fujunwei:
@darktears  @rakuco  do you have any more comments?

darktears:
It removes the content.gyp dependency, that's ok. Can we avoid to remove the file so we don't get conflicts when rebasing?

I don't think https://github.com/fujunwei/crosswalk/commit/3395073e0a3f9fad34bf6270523add6b82e1a15d should land. Just roll the DEPS.

fujunwei:
Thanks for your propose. it's a good approach. I will refine it.

fujunwei:
My approach is to move the depended file to out/Debug before compiling the content java source code, but customer have to be careful it doesn't commit the modification.
Do you think so?

fujunwei:
I will move it back after the process of it is  complete.

baleboy:
@rakuco @darktears any further comments?

darktears:
"My approach is to move the depended file to out/Debug before compiling the content java source code, but customer have to be careful it doesn't commit the modification.
Do you think so? " -> I don't understand what you mean.

darktears:
Why you don't try to compile out :
- browser/android/background_sync_launcher_android.cc
- browser/android/background_sync_launcher_android.h
- public/android/java/src/org/chromium/content/browser/BackgroundSyncLauncher.java

In their respective .gyp/gpyi files. It should be more than enough. Don't remove the files from the tree (so we don't have conflicts at rebase) we just don't build them.

What would not make that approach to work?

darktears:
You probably need to make sure this part is ignored (it seems to be the only call site of the added feature) : https://codereview.chromium.org/1140813006/patch/340001/350008 . Sounds like you can just remove what's inside the #ifdef Android which will limit the conflicts greatly (keeps the rest of the code).

fujunwei:
Because only BackgroundSyncLauncherService.java depend on android library support, so we just don't build the file.
My approach is moving the BackgroundSyncLauncherService.java to out/Debug before compiling the content java source code, and move it back after finishing.

public/android/java/src/org/chromium/content/browser/BackgroundSyncLauncherService.java

fujunwei:
I don't understand why do we  try to compile those three file out? 
browser/android/background_sync_launcher_android.cc
browser/android/background_sync_launcher_android.h
public/android/java/src/org/chromium/content/browser/BackgroundSyncLauncher.java

fujunwei:
@darktears PTAL. 
Relative PR is https://github.com/crosswalk-project/crosswalk/pull/3291

fujunwei:
ping @darktears

rakuco:
I've been trying to reproduce this issue locally, and learning some basic Gradle in the process, in order to be able to have a more in-depth discussion.

I won't get into this specific fix at the moment; I do not like the way the `find(1)` call was hacked and how non-obvious the patch is, but if I understood everything correctly maybe we shouldn't try to remove the v13 dependency at all in the first place: right now there's only this specific `.java` file that depends on v13, but nothing prevents other parts of the Chromium stack from using it in the future, and then we'll be playing whack'a'mole trying to disable things just to remove this dependency.

If I also understood it correctly, we also cannot try hacking our AAR file since one might use Crosswalk in a project that does not pull in v4 APIs and we end up with a project that does not build. Is that correct?

In case I got everything right, would it be a solution to change `xwalk.gradle` in the cordova-plugin-crosswalk-webview repository and do something like

```
configurations {
  all*.exclude group: 'com.android.support'
}
```

selectively, only if it is being pulled as a dependency (I'm not sure if it is possible to check other dependencies like this)? Is there something similar for Eclipse? I started a new Android Studio project here and it did not include support-v4 by default, as far as I can see.

Another possibility would be updating our documentation and mentioning this issue to users, but I don't like this idea very much (things won't "Just Work").

The bottomline is that I'd like to avoid having to change chromium-crosswalk here: following the idea in this patch is not future-proof.

rakuco:
One other thing that might work: is it possible to remove those support-v* libraries from the classes.jar in the AAR file, and then we list them as dependencies in the POM file in Maven? The downside is that we'd then need to take care of maintaining this dependency list.

fujunwei:
Do you mean exclude the classes of android support when packaging xwalk library jar?
Similar PR is https://github.com/fujunwei/crosswalk/commit/18a5a6c003380c048858992e65ca4f06a6d05e1c

rakuco:
Yes, that's one of the things I suggested. However, as I said above, doesn't that break things if one does not explicitly depend on the Android support classes now?

r0b5t4:
@rakuco I think build-time ("API" compat) is not as big a concern as with traditional shared libraries. The common approach is to rebuild your package when moving to a newer version. (Except maybe for shared crosswalk)

rakuco:
But in this case, isn't "rebuilding the package" going to fail? My understanding was that if one has an app that only depends/uses Crosswalk 15 in webview mode and moves to Crosswalk 16 the build won't work unless the person also adds a dependency on the android-support-v4 package.

r0b5t4:
My limited understanding of Android Studio suggests that if you import a new crosswalk (aar), gradle will do its thing, and resolve the dependencies. I don't know of any way you could update crosswalk without going through that. I was assuming the support library would be declared as a dependency by the aar.

fujunwei:
@rakuco Yes, it doesn't break things.
Please help to take a look the refined PR https://github.com/crosswalk-project/crosswalk/pull/3291

baleboy:
All, our cordova testing on 16 and 17 has been blocked for a month. I'm all for finding the best solution, but can we take whatever fix we have now to unblock the testing and figure out the best way forward without the pressure of releasing?

fujunwei:
@baleboy The Cordova testing can be tested and doesn't depend on this PR.

baleboy:
I see, I misunderstood the blocker status

fujunwei:
Because the latest cca testing doesn't use the cordova-plugin-chrome-apps-notifications which depend on android support library. But the issue still exist.
Anyway we expect it can be merged as soon as possible.

PR #283, opened 2015-09-29
[Lite] Remove "-fstack-protector" compile option

heke123:
-fstack-protector is used for preventing buffer overflow attack.
But it also brings larger binary size. it is also controlled by
flag: "use_optimize_for_size_compile_option".

With lzma, ~200K binary size reduced for EmbeddedShell.apk.


heke123:
@halton  @PeterWangIntel please help to review. thanks.

PeterWangIntel:
LGTM

PR #285, opened 2015-10-09
Make the HTML tag "video" unknown for WebKit if WebVideo disabled

PeterWangIntel:
I've awared of this problem during investigation of XWALK-5147. It caused by user trying to access a webpage with complicated 'video' tag related content when WebVideo is disabled.

Although the calling of system Media Player is excluded when disable
WebVideo, if the WebKit still takes <video> as a known tag, there will
be risk of many unexpected problems.

### DIFF
--- a/content/child/runtime_features.cc
+++ b/content/child/runtime_features.cc
@@ -228,6 +228,11 @@ void SetRuntimeFeaturesDefaultsAndUpdateFromArgs(
           blink::WebString::fromLatin1(feature), false);
     }
   }
+
+// Make the tags of  disabled features unknown for WebKit
+#ifdef DISABLE_WEB_VIDEO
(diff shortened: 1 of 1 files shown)


PeterWangIntel:
@halton @heke

halton:
@PeterWangIntel code lgtm, I would like to wait u enhance comment with `BUG=XWALK-5147` before merging.

PeterWangIntel:
@halton 
Please help to merge this patch.

PeterWangIntel:
It's resolve the crash when 'Web Video' being disabled.
I'll roll the DEPS.xwalk then.

halton:
Please roll

PeterWangIntel:
thx

PR #293, opened 2015-11-03
[Android] Fix crash in unit test

lincsoon:
The change of VSyncMonitor in the commit af926cfcaed81686e6dbf782932b22deed1bef4c
causes crash of XWalkCoreShell and XWalkCoreInternallShell,
and then blocks a lot of test cases.

This commit is to fix the crash.

Related bug: XWALK-5241

### DIFF
--- a/ui/android/java/src/org/chromium/ui/VSyncMonitor.java
+++ b/ui/android/java/src/org/chromium/ui/VSyncMonitor.java
@@ -96,6 +96,7 @@ public class VSyncMonitor {
                 @Override
                 public void doFrame(long frameTimeNanos) {
                     TraceEvent.begin("VSync");
+                    mHandler.removeCallbacks(mSyntheticVSyncRunnable);
                     if (useEstimatedRefreshPeriod && mConsecutiveVSync) {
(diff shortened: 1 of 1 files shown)


halton:
@lincsoon Seems upstream made same change on first part, https://code.google.com/p/chromium/codesearch#chromium/src/ui/android/java/src/org/chromium/ui/VSyncMonitor.java&l=74. But I do not find the second part.

Please consider to use upstream change or make a [Backport] label commit.

rakuco:
This looks like the exact same thing I had to fix for the M45 rebase, it's surprising that it happened again. The problem is that we're reverting a ton of commits to reintroduce support for ICS devices, and meanwhile newer commits have been introduced upstream to the same files we're touching, and if one is not careful we end up with bad merges that cause problems like this.

rakuco:
lgtm, btw. `diff` says `VSyncMonitor.java` looks identical to the version in master, which is a good sign. We need to think of a way of not having this problem on every rebase though.

rakuco:
since you mentioned via email that this change fixes many crashes but not all, I've edited the pull request message a bit so that the corresponding JIRA ticket isn't closed when I merge this PR.

PR #294, opened 2015-11-03
[Android] Stop using org.apache.http classes in AccessibilityInjector.

rakuco:
As part of the efforts to stop depending on external packages in the
code, replace usage of org.apache.http classes with an implementation
that only uses apache.net.Uri.

This allows us to drop the dependency on the legacy_http_client target
and end up having to bundle org.apache.http.legacy.jar into
xwalk_core_library.jar.

Not upstreamable, as the code in question was removed upstream when they
dropped support for ICS.

Related to: XWALK-5092

### DIFF
--- a/content/content.gyp
+++ b/content/content.gyp
@@ -599,7 +599,6 @@
           'dependencies': [
             'java_set_jni_headers',
             'motionevent_jni_headers',
-            '../third_party/android_tools/android_tools.gyp:legacy_http_javalib',
           ],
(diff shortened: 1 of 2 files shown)


rakuco:
ping @darktears @lincsoon

halton:
LGTM, this is perfect solution, thanks for @rakuco driving this.

fujunwei:
LGTM.

PR #297, opened 2015-11-10
Revert "To disable Plugin"

PeterWangIntel:
This reverts commit e6c0772362385cf059b15e6e71a6603df636701e.


PeterWangIntel:
@halton 
Revert to unify the build flag

PR #298, opened 2015-11-10
[Lite] To disable Plugins

PeterWangIntel:
Flag = enable_plugins
Reduced size = 440K


PeterWangIntel:
@halton, It's been verified, thx.

PR #299, opened 2015-11-12
crosswalk-16: [Android] Stop using org.apache.http classes in AccessibilityInjector.

rakuco:
As part of the efforts to stop depending on external packages in the
code, replace usage of org.apache.http classes with an implementation
that only uses apache.net.Uri.

This allows us to drop the dependency on the legacy_http_client target
and end up having to bundle org.apache.http.legacy.jar into
xwalk_core_library.jar.

Not upstreamable, as the code in question was removed upstream when they
dropped support for ICS.

Related to: XWALK-5092

(cherry picked from commit 5909ede1bf2c0d7e05b2ea1cd9542fba4ae8e0dd)

### DIFF
--- a/content/content.gyp
+++ b/content/content.gyp
@@ -599,7 +599,6 @@
           'dependencies': [
             'java_set_jni_headers',
             'motionevent_jni_headers',
-            '../third_party/android_tools/android_tools.gyp:legacy_http_javalib',
           ],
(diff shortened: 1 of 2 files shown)


rakuco:
ping @halton @lincsoon @darktears

halton:
LGTM

PR #301, opened 2015-11-17
[Android] Fix crash in unit test

fujunwei:
Load data when the scheme of url is data:.
The test case is ShouldOverrideUrlLoadingTest#testDoesNotCauseOnReceivedError
in XWalkCoreTest.

BUG=XWALK-5241

### DIFF
--- a/content/renderer/render_frame_impl.cc
+++ b/content/renderer/render_frame_impl.cc
@@ -4612,9 +4612,9 @@ void RenderFrameImpl::NavigateInternal(
     base::TimeTicks renderer_navigation_start = base::TimeTicks::Now();
 
     // Perform a navigation to a data url if needed.
-    if (!common_params.base_url_for_data_url.is_empty() ||
-        (browser_side_navigation &&
-         common_params.url.SchemeIs(url::kDataScheme))) {
+    if ((!common_params.base_url_for_data_url.is_empty() ||
+         browser_side_navigation) &&
+        common_params.url.SchemeIs(url::kDataScheme)) {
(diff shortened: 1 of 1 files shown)


fujunwei:
@lincsoon @rakuco PTAL

rakuco:
This pull request:

* Does not point to the upstream change that caused the problem
* Does not say if this change is temporary and whether it also affects upstream (and, in this case, whether the fix was also sent to Chromium)
* Does not say why the behavior in Crosswalk needs to be different from what's upstream and what the problem actually is.

Therefore I cannot review this.

darktears:
Please add details. Changing the order of the condition seems sneaky to me and will probably affect a lot more.

fujunwei:
Further investigation about the reason of this changes in Chromium upstream. It's  discussed in https://code.google.com/p/chromium/issues/detail?id=522567.
Because the next branch is base on Chromium 46.0.2490.33 that don't include this fixed commit, so i will cherry pick it to refine the PR.

The latest Chromium 46.0.2490.33 in next branch is https://github.com/crosswalk-project/chromium-crosswalk/commit/185543c5b54794ebbdd3f7732e480eb4a2a11648.

baleboy:
or should we just update the next branch to the latest M46 chromium? Eventually we'll have to do it anyway.

rakuco:
I agree, I think we can just move to a more recent M46, which means XWALK-5139 becomes urgent.

fujunwei:
Ok, I will close the PR.

PR #304, opened 2016-01-04
[Android] Expose the constructor of ContentView

lincsoon:
Unlike WebView, XWalkView uses a ContentView object as its child view,
which is actually receiving all kinds of events such as touching and
clicking. To let the XWalkView object receive these events too, it's
necessary to override relevant methods of ContentView and redirect
them to the XWalkView object. So we add XWalkContentView and have it
extend ContentView. To make this possible, the contructor of ContentView
must be public.

XWalkContentView is a temporary solution. We are targeting to optimize
the hierarchy of XWalkView to make it more like WebView. This task can
be tracked via XWALK-6118.

BUG=XWALK-6014

### DIFF
--- a/content/public/android/java/src/org/chromium/content/browser/ContentView.java
+++ b/content/public/android/java/src/org/chromium/content/browser/ContentView.java
@@ -56,7 +56,7 @@ public class ContentView extends FrameLayout
      *                access the current theme, resources, etc.
      * @param cvc A pointer to the content view core managing this content view.
      */
-    ContentView(Context context, ContentViewCore cvc) {
+    public ContentView(Context context, ContentViewCore cvc) {
         super(context, null, android.R.attr.webViewStyle);
 
         if (getScrollBarStyle() == View.SCROLLBARS_INSIDE_OVERLAY) {


lincsoon:
@rakuco Please help to review

rakuco:
In case the other approach I suggested in the related Crosswalk pull request is not feasible: have you tried upstreaming this? We always try the upstream first approach.

rakuco:
It's even valid to say in your CL that this makes the behavior similar to what we had before in Chromium, when the constructor was `protected`.

lincsoon:
I agree your approach too and actually other colleagues are working on it. But it's a long term issue that getting rid of `ContentView`, we must fix it carefully to avoid potential problems. So currently, I this this pull request is the most effective way.

rakuco:
Is there a tracking bug for these changes you mention, and/or an ETA and target release? Getting rid of the `ContentView` child class looked somewhat trivial at first sight, but I haven't investigated too deeply.

I'd just like to have some commitment that this will be addressed soon and we won't be forced to carry this patch and the Crosswalk one forever as it has happened before.

lincsoon:
I filed a ticket(https://crosswalk-project.org/jira/browse/XWALK-6118) to track this issue. But I think the change is not going to happen soon.

rakuco:
That's a bit disappointing; I've added some comments to the JIRA ticket to make it clear that this should have a higher priority.
Please at least mention XWALK-6118 in this commit and make it clear in the commit message that this is supposed to be a temporary solution.

lincsoon:
Done! Please merge this first so that I can update the dependency file of https://github.com/crosswalk-project/crosswalk/pull/3472

PR #305, opened 2016-01-06
[Windows] Support RealSense Cameras

huningxin:
Add Intel RSSDK based video capture device VideoCaptureDeviceRSWin.

It allows:
1. Crosswalk apps are able to share the RealSense camera with other
   native apps simultaneously.
2. Crosswalk apps are able to use getUserMedia API to preview RealSense
   camera in real-time while accessing RSSDK middlewares in Crosswalk
   extensions simultaneously.

This feature is behind --use-rs-video-capture flag.

TEST=
On a device with RealSense camera and RSSDK, execute
1. DF_RawStreams.exe (RSSDK sample)
2. xwalk.exe --use-rs-video-capture
   https://webrtc.github.io/samples/src/content/devices/input-output/
Crosswalk and DF_RawStreams.exe should capture video frame from
RealSense camera simultaneously.

BUG=XWALK-6028

### DIFF
--- a/media/base/media_switches.cc
+++ b/media/base/media_switches.cc
@@ -78,6 +78,9 @@ const char kForceWaveAudio[] = "force-wave-audio";
 // http://crbug.com/259165 for more details.
 const char kTrySupportedChannelLayouts[] = "try-supported-channel-layouts";
 
+// Use RSSDK for video capture.
+const char kUseRsVideoCapture[] = "use-rs-video-capture";
+
 // Number of buffers to use for WaveOut.
 const char kWaveOutBuffers[] = "waveout-buffers";
 #endif
--- a/media/capture/video/win/video_capture_device_factory_win.cc
+++ b/media/capture/video/win/video_capture_device_factory_win.cc
@@ -19,6 +19,11 @@
 #include "media/base/media_switches.h"
 #include "media/base/win/mf_initializer.h"
 #include "media/capture/video/win/video_capture_device_mf_win.h"
+#if defined(USE_RSSDK)
+#include "media/capture/video/win/video_capture_device_rs_win.h"
+#else
+#include "media/capture/video/win/video_capture_device_rs_win_null.h"
+#endif
 #include "media/capture/video/win/video_capture_device_win.h"
 
 using base::win::ScopedCoMem;
@@ -377,6 +382,8 @@ VideoCaptureDeviceFactoryWin::VideoCaptureDeviceFactoryWin() {
   // can also be forced if appropriate flag is set and we are in Windows 7 or
   // 8 in non-Metro mode.
   const base::CommandLine* cmd_line = base::CommandLine::ForCurrentProcess();
+  use_rssdk_ = cmd_line->HasSwitch(switches::kUseRsVideoCapture) &&
+      VideoCaptureDeviceRSWin::IsSupported();
   use_media_foundation_ =
       (base::win::IsMetroProcess() &&
        !cmd_line->HasSwitch(switches::kForceDirectShowVideoCapture)) ||
@@ -388,7 +395,13 @@ scoped_ptr<VideoCaptureDevice> VideoCaptureDeviceFactoryWin::Create(
     const Name& device_name) {
   DCHECK(thread_checker_.CalledOnValidThread());
   scoped_ptr<VideoCaptureDevice> device;
-  if (device_name.capture_api_type() == Name::MEDIA_FOUNDATION) {
+  if (device_name.capture_api_type() == Name::RSSDK) {
+    DCHECK(device_name.capture_api_type() == Name::RSSDK);
+    device.reset(new VideoCaptureDeviceRSWin(device_name));
+    DVLOG(1) << " RSSDK Device: " << device_name.name();
+    if (!static_cast<VideoCaptureDeviceRSWin*>(device.get())->Init())
+      device.reset();
+  } else if (device_name.capture_api_type() == Name::MEDIA_FOUNDATION) {
     DCHECK(PlatformSupportsMediaFoundation());
     device.reset(new VideoCaptureDeviceMFWin(device_name));
     DVLOG(1) << " MediaFoundation Device: " << device_name.name();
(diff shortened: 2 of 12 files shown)


huningxin:
@darktears @kenchris @halton , PTAL. Thanks!

rakuco:
Given the amount of code in this patch, I have to ask the obligatory questions about the upstreaming plans for this. It'd really suck if we were bound to carry this amount of code ourselves forever.

huningxin:
@rakuco , thanks for bringing up, this is a must-answer question. 

The short answer is the upstream is definitely the next step.

Some details:

There is my previous CL to Chromium: Add Intel Perceptual Computing SDK based depth video capture for Windows (https://codereview.chromium.org/262383003/) 
 
P.S. The Perceptual Computing SDK in the Chromium CL is ancient name of RealSense SDK. (I will use RSSDK alone in following texts)

In that CL, the reviewer, mcasas, supported the approach which adds a new VideoCaptureDevice implementation via RSSDK API. His concern was about the RSSDK dependency at build time. The solution is to pull the RSSDK headers and library loader into third_party. So Chromium can build without RSSDK installation. However, that solution was blocked by open source headers and library loader of RSSDK were not available at that time.

In this PR, with the `VideoCaptureDeviceRSWin` and the `libpxc`, the implementation exactly follows the direction and solve the blocking issue of that CL. It would be a good base to resume the upstream.

This PR is a base work for two use cases: 1) exposing RSSDK middlewares through Crosswalk Extensions; 2) exposing raw depth streaming via W3C Media Capture Depth Stream Extensions API (http://w3c.github.io/mediacapture-depth/). 

For Crosswalk, by landing this PR, it allows developers to experiment use case 1. 

For upstream, the use case 2 would be a good driver to push that CL, I will focus on that.

Given devices and applications for RealSense cameras are getting mainstream, I believe there is a good reason to adopt this feature, by Crosswalk first, and by upstream eventually .

Any suggestions?

rakuco:
Thanks for the detailed background, some of it could even go into the commit message.

Is there anyone currently working on resuming the upstreaming of the code? For something as big as this, I see it as "must do", not "would be nice to do some day". Another concern I have related to this is that the original CL review suggested the glue approach used for several OS X libraries: instead of linking against a library, you try to load it dynamically at runtime and clients check if it worked before using your glue code. The upside is that everything Just Works (TM) with and without the libraries being present; downsides include an additional layer of complexity in the glue code and the fact that the library must be present on the system for things to work. This PR, on the other hand, builds libpxc and links against it instead. It is a valid approach, I'm just curious about how what approach you intend to take upstream.

media/base/media_switches.cc
```diff
@@ -78,6 +78,9 @@ const char kForceWaveAudio[] = "force-wave-audio";
 // http://crbug.com/259165 for more details.
 const char kTrySupportedChannelLayouts[] = "try-supported-channel-layouts";
 
+// Use RSSDK for video capture.
+const char kUseRsVideoCapture[] = "use-rs-video-capture";
```
rakuco: I think you'll also need a build system flag in addition to the runtime command line one: as is, the code will break the content shell builds, as libpxc is pulled in by Crosswalk, which is not checked out in the content shell bots.

media/capture/video/win/video_capture_device_factory_win.cc
```diff
@@ -377,6 +378,8 @@ VideoCaptureDeviceFactoryWin::VideoCaptureDeviceFactoryWin() {
   // can also be forced if appropriate flag is set and we are in Windows 7 or
   // 8 in non-Metro mode.
   const base::CommandLine* cmd_line = base::CommandLine::ForCurrentProcess();
+  use_rssdk_ = cmd_line->HasSwitch(switches::kUseRsVideoCapture) &&
+      VideoCaptureDeviceRSWin::IsSupported();
```
huningxin: @rakuco, this checks whether valid RSSDK library is present or not.

huningxin:
> This PR, on the other hand, builds libpxc and links against it instead. It is a valid approach, I'm just curious about how what approach you intend to take upstream.

The libpxc is exactly the glue approach. It loads the dynamic library of RSSDK at runtime. The `VideoCaptureDeviceRSWin::IsSupported()` (I mentioned inline) calls libpxc to check whether RSSDK is present. It returns true if RSSDK is present and can be loaded, otherwise it will return false.

I will use this approach to upstream.

media/base/media_switches.cc
huningxin: Thanks for letting me know this. I will add the build flag.

huningxin:
@rakuco , the build flag is added in latest PR update. Please take another look. Thanks!

huningxin:
ping @rakuco and @darktears , Thanks!

rakuco:
> The libpxc is exactly the glue approach. It loads the dynamic library of RSSDK at runtime. The VideoCaptureDeviceRSWin::IsSupported() (I mentioned inline) calls libpxc to check whether RSSDK is present. It returns true if RSSDK is present and loaded successfully, otherwise it will return false.

Oh, I see, stupid me :-) I originally thought you were linking against the SDK in libpxc.

In this case, I think you can simplify the build-time checks (which should be temporary until Chromium at least starts fetching libpxc) so that you don't end up with a lot of ugly checks scattered around:
* Add `capture/video/win/video_capture_device_rs_win_null.{cc,h}` which are built if `use_rssdk=0`. It can contain just the parts of `VideoCaptureDeviceRSWin` which are called from common code in `media`. For example, its implementation of `VideoCaptureDeviceRSWin::IsSupported()` always returns `false` etc. In other words, in one case you link against libpxc and build the real code, and in the other you just build the stubs.
* Remove the `#if defined(USE_RSSDK)` checks and add some additional checks for `VCDRSW::IsSupported()` if necessary.

huningxin:
@rakuco , I like the `video_capture_device_rs_win_null.{cc,h}` idea as well. Updated the PR to implement that, please take another look. Thanks!

rakuco:
great, lgtm!

rakuco:
please remember to roll `DEPS.xwalk`

PR #309, opened 2016-01-21
[Windows] Maximize the window over the taskbar in the fullscreen mode.

joone:
It allows the users to close the fullscreen application.

BUG=XWALK-4861


joone:
This CL works with https://github.com/crosswalk-project/crosswalk/pull/3497

PR #311, opened 2016-01-22
[Windows] Maximize the window over the taskbar in the fullscreen mode.

joone:
It allows the users to close the fullscreen application.

BUG=XWALK-4861


joone:
This CL needs https://github.com/crosswalk-project/crosswalk/pull/3501

rakuco:
Why not just update #309?

joone:
@rakuco  I deleted the branch and created the new branch with the same name, but the new pull request was created like this.

darktears:
@pozdnyakov what do you think? It can be a compromise until we integrate better with the VM (which would require a lot more work on the views toolkit).

joone:
This is a temporary solution. Metro driver will be used to make
Crosswalk applications work like Windows universal application,
which is in src/win8/metro_driver of Chromium source code.

pozdnyakov:
@joone @darktears but metro driver is a bit different case, right? For chromium it just provides CrOS session, how would it help with Crosswalk fullscreen apps?

PR #312, opened 2016-01-25
[Android][ICU] Fix error always true of icu_utils

heke123:
[Android][ICU] Fix error always true of icu_utils

icu_utils.[cc|h] is the ICU JNI alternative C++ bridge, it will be enabled
by build flag "use_icu_alternatives_on_android=1". The in/out parameter *error
of functions toLower(), toUpper() and normalize() are be set true at any time.
The caller will detect that error and then drop the result.

The fix simply to set the error as false. Even the conversion failed, the
original string will be returned, which is no harm.

BUG=XWALK-6161

### DIFF
--- a/base/icu_alternatives_on_android/icu_utils.cc
+++ b/base/icu_alternatives_on_android/icu_utils.cc
@@ -243,7 +243,7 @@ int toLower(UChar* result, int resultLength, const UChar* src, int srcLength, bo
       Java_IcuUtils_toLowerString(env, sourceString.obj());
 
   string16 resultString = ConvertJavaStringToUTF16(java_result);
-  *error = true;
+  *error = false;
   return copy(result, resultLength, resultString.data(), resultString.size());
 }
 
@@ -268,7 +268,7 @@ int toUpper(UChar* result, int resultLength, const UChar* src, int srcLength, bo
       Java_IcuUtils_toUpperString(env, sourceString.obj());
 
   string16 resultString = ConvertJavaStringToUTF16(java_result);
-  *error = true;
+  *error = false;
   return copy(result, resultLength, resultString.data(), resultString.size());
 }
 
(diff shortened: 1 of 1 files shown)


heke123:
'\u0130mage/png' (it is İmage/png)=> 'image/png'.  the func toLower() is implemented in java side.
In our c++ bridge layer, it is wrong to always set "error=false". Because some caller will drop the result from tolower if they find "error = true".
if the convert is really failed in java side, java doc show it will return original src string itself.  it is still safe.

heke123:
@halton  @PeterWangIntel

halton:
Code LGTM, please make a nicer git log. You first comment about the reason should be added in abstract to explain, and Bug=XWalk-6161 -> BUG=XWALK-6161

heke123:
@halton  thanks for your reminder. I updated the git log.

halton:
Still not nice as I am reading, even some grammar errors and typos (for eg: implemenation->implementation). So, how about below?
```
[Android] [ICU] Fix error always true of icu_utils

icu_utils.[cc|h] is the ICU JNI alternative C++ bridge, it will be enabled
by build flag "use_icu_alternatives_on_android=1". The in/out parameter *error
of functions toLower(), toUpper() and normalize() are be set true at any time.
The caller will detect that error and then drop the result.

The fix simply to set the error as false. Even the conversion failed, the
original string will be returned, which is no harm.

BUG=XWALK-6161
```

heke123:
@halton  your comments are really much better. I'll improve my comments in following commits.
Thanks very much.
I have updated the git log as your suggest.

halton:
Please change the PR title as well, then i'll merge.

heke123:
@halton  PR title changed.

halton:
LGTM and thanks for your instant reworking.

halton:
@heke123 please remember update DEPS.xwalk

PR #313, opened 2016-01-25
XWALK-4992 [windows] App does not support gamepad

astojilj:
XWALK-4992 [windows] App does not support gamepad

    While chromium is bundling xinput1_3.dll from DirectX redist, using whatever
    platform suports. Logic about what dll is in use for different platforms is
    copied from Xinput.h.
    In order to target all of the Windows versions with the same binary, chosen to
    keep dynamic linking approach. Otherwise, there would be a problem with
    XInputEnable not available in xinput dll before Windows8 and deprecated on
    Windows 10.

    Snippet from <ProgramFiles>\Windows Kits\10\Include\10.0.10586.0\um\Xinput.h:

    #if(_WIN32_WINNT >= _WIN32_WINNT_WIN8)
    #define XINPUT_DLL_A  "xinput1_4.dll"
    #define XINPUT_DLL_W L"xinput1_4.dll"
    #else
    #define XINPUT_DLL_A  "xinput9_1_0.dll"
    #define XINPUT_DLL_W L"xinput9_1_0.dll"
    #endif

    For detailed information on Windows library loading and search path priorities,
    check https://msdn.microsoft.com documentation on LoadLibrary().

### DIFF
--- a/chrome/installer/mini_installer/chrome.release
+++ b/chrome/installer/mini_installer/chrome.release
@@ -39,7 +39,6 @@ natives_blob.bin: %(VersionDir)s\
 resources.pak: %(VersionDir)s\
 snapshot_blob.bin: %(VersionDir)s\
 syzyasan_rtl.dll: %(VersionDir)s\
-xinput1_3.dll: %(VersionDir)s\
 #
 # Sub directories living in the version dir
 #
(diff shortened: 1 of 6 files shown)


astojilj:
@darktears 
Could you please review it? Thanks.

darktears:
It looks fine to me. Can you clarify (and in the commit message) how the DLL are supposed to be found in the first place? I get that you special case them per platform but how the loading works?

darktears:
LGTM, make sure to roll the DEPS

astojilj:
@darktears 
Updated the description. Still didn't test this on other versions except on Windows 10. Notifying when ready.

astojilj:
@darktears 
Verified also on Wind8. OK to merge.

PR #314, opened 2016-01-28
[Blink] WebCL: Only set ENABLE_WEBCL in the relevant files.

rakuco:
This change (together with a follow-up in the Crosswalk repository
itself) moves the definition of the ENABLE_WEBCL macro to Blink's
features.gypi, so that it is only passed to the relevant modules and
bindings files intead of all the 15000+ files part of Crosswalk's build.

This part of the fix only adds the definition to the Blink side; the
Crosswalk patch will stop defining the macro for the other files.


rakuco:
@chunkyan @darktears

shaoboyan:
LGTM, this is part of my rcode to fix XWALK-6200

PR #317, opened 2016-02-17
[Android][ICU] Add Shift-JIS codec support.

heke123:
The "use_icu_alternatives_on_android=1" flag cut off the icu c++
library, which breaks the Shift-JIS encode and decode functions.
So we add a TextCodecICUAlternatives class which provides such
functions by using the Java class CharsetEncoder and
CharsetDecoder.

BUG=XWALK-6175

### DIFF
--- a/third_party/WebKit/Source/wtf/text/TextCodecICUAlternatives.cpp
+++ b/third_party/WebKit/Source/wtf/text/TextCodecICUAlternatives.cpp
@@ -1,0 +1,69 @@
+// Copyright 2014 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "config.h"
+#include "wtf/text/TextCodecICUAlternatives.h"
+#include "wtf/PassOwnPtr.h"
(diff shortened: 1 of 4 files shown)


heke123:
@PeterWangIntel @halton  Please help review. thanks.

PeterWangIntel:
As face to face review, LGTM.

third_party/WebKit/Source/wtf/text/TextCodecICUAlternatives.cpp
```diff
@@ -0,0 +1,69 @@
...
+namespace WTF {
+
+TextCodecICUAlternatives::TextCodecICUAlternatives(const TextEncoding& encoding)
+    : m_encoding(encoding)
+{
+}
+
+void TextCodecICUAlternatives::registerEncodingNames(EncodingNameRegistrar registrar)
+{
+    //TODO, add more charsets supported by Java.
```
heke123: In Java side, theoretically it can support lots of charset.  This error is caused by decoding Shift-JIS, so I just register Shift-JIS here.  we can add other charset if necessary in future.

PR #318, opened 2016-02-17
[Lite] [WORKAROUND] One assert always blocks debug

heke123:
This Assert() blocks debugging. Only happens in debugging build,
not in release build. So just remove it.

It is just a workaround for debugging convenience, we won't port
it to next release version.

### DIFF
--- a/gpu/command_buffer/service/program_manager.cc
+++ b/gpu/command_buffer/service/program_manager.cc
@@ -1107,7 +1107,7 @@ bool Program::DetectShaderVersionMismatch() const {
         return true;
       }
       version = shader->shader_version();
-      DCHECK(version != Shader::kUndefinedShaderVersion);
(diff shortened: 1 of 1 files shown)


halton:
Does this debug build failure happen on master branch? If so, we need fix on master first and then cherry-pick

heke123:
@halton, It is not a build failure, it happens in runtime, I don't know what happens in master branch, but everytime before I debug lite, i have to apply this modify first..

heke123:
Actually it is just for debugging convenience.

halton:
> @halton, It is not a build failure, it happens in runtime, I don't know what happens in master branch, but everytime before I debug lite, i have to apply this modify first..

Thanks for clarification. From the code, I guess this assertion should happen on master debug build as well, we need fix it on master first.

heke123:
@halton  OK, Understand now.

PR #320, opened 2016-02-25
Avoid previous hack in gyp files for conditional compilation of WebCL

hujiajie:
This avoids previous hack in gyp files for conditional compilation of WebCL. WebCL is built for all platforms now, but it will be disabled during runtime for platforms other than Android.


hujiajie:
@shaoboyan, PTAL, thanks

shaoboyan:
LGTM and finally could remove these tricks !

hujiajie:
@Bysmyyr, please help to check if the patch will break the Android / Windows / Linux build. Thanks

third_party/WebKit/Source/modules/webcl/WebCLOpenCL.cpp
```diff
@@ -197,9 +200,14 @@ cl_int (CL_API_CALL *web_clGetGLTextureInfo)(cl_mem, cl_gl_texture_info, size_t,
 #define clEnqueueBarrierWithWaitList web_clEnqueueBarrierWithWaitList
 #define clCreateFromGLTexture web_clCreateFromGLTexture
 
-#if OS(LINUX) || OS(ANDROID)
+#if OS(ANDROID)
 #define MAP_FUNC(fn)  { *(void**)(&fn) = dlsym(handle, #fn); }
 #define MAP_FUNC_OR_BAIL(fn)  { *(void**)(&fn) = dlsym(handle, #fn); if(!fn) return false; }
+#else
+#define MAP_FUNC(fn) ASSERT_NOT_REACHED();
```
rakuco: Isn't this problematic in the sense that `WebCL.cpp` will call `init()`, and on !Android this will result in a lot of calls to `ASSERT_NOT_REACHED()` that will crash Crosswalk if assertions are enabled?

hujiajie: On !Android MAP_FUNC() and its friends should never be reached. init() should return immediately after getCLHandle().

rakuco: I see, thanks. I guess we're ready to merge. This looks very nice, I guess we can also remove the `use_webcl` gyp flag from Crosswalk?

rakuco:
FYI, I had to push an additional commit adding the `CORE_EXPORT` macro to `WebCLException`'s declaration, otherwise the symbol wouldn't be exported and linking with `component=shared_library` would fail. I think this bug has been present ever since I moved `WebCLException` to `core/`, but it only manifested itself now that the WebCL code is built on other platforms where it is more common to build with `component=shared_library` like Linux.

hujiajie:
LGTM, thanks

PR #321, opened 2016-02-27
Page Visibility API: visibilitychange on Windows screen lock/unlock

astojilj:
BUG=XWALK-5503

Some OSs (e.g. Windows and Linux) don't mark windows as hidden on screen
lock or when other opaque windows fully cover them.
Note that e.g. OSX and Android do this while Windows and Linux don't.
OnSoftVisibilityChanged enables that OS window state is kept as is (on
Linux and Windows) and that e.g. power saving PageVisibility API is still
properly triggered.

The first patch here targets Windows and screen lock.

### DIFF
--- a/ui/views/controls/native/native_view_host.cc
+++ b/ui/views/controls/native/native_view_host.cc
@@ -127,6 +127,22 @@ void NativeViewHost::VisibilityChanged(View* starting_from, bool is_visible) {
   Layout();
 }
 
+void NativeViewHost::OnSoftVisibilityChanged(bool visible) {
+  if (!native_view_ || !native_wrapper_.get())
+    return;
+
+  if (GetVisibleBounds().IsEmpty())
+    return;
+
+  // If the widget wasn't .e.g. minimized, hide and show would trigger
+  // PageVisibility API, enabling JS to save power while screen is locked.
+  if (visible) {
+    Layout();
+  } else {
+    native_wrapper_->HideWidget();
+  }
+}
+
 bool NativeViewHost::GetNeedsNotificationWhenVisibleBoundsChange() const {
   // The native widget is placed relative to the root. As such, we need to
   // know when the position of any ancestor changes, or our visibility relative
--- a/ui/views/view.cc
+++ b/ui/views/view.cc
@@ -1840,6 +1840,12 @@ void View::PropagateNativeThemeChanged(const ui::NativeTheme* theme) {
   OnNativeThemeChanged(theme);
 }
 
+void View::PropagateSoftVisibilityChanged(bool visible) {
+  for (int i = 0, count = child_count(); i < count; ++i)
+    child_at(i)->PropagateSoftVisibilityChanged(visible);
+  OnSoftVisibilityChanged(visible);
+}
+
 // Size and disposition --------------------------------------------------------
 
 void View::PropagateVisibilityNotifications(View* start, bool is_visible) {
--- a/ui/views/view.h
+++ b/ui/views/view.h
@@ -1073,6 +1073,8 @@ class VIEWS_EXPORT View : public ui::LayerDelegate,
   // FocusManager manages this view.
   virtual void NativeViewHierarchyChanged();
 
+  virtual void OnSoftVisibilityChanged(bool visible) {}
+
   // Painting ------------------------------------------------------------------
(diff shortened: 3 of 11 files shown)


astojilj:
@darktears could you please review? I'll post it to upstream review too.

astojilj:
@darktears  Updated the signatures to make them more common. Patch is still handling screen lock only. Could you please review this? Thanks.

darktears:
@astojilj that's ok to land it here but do you think that Google would be interested to have the patch upstream. It sound legit to me that when I lock the screen the app get the event and can pause a game say.

@pozdnyakov any other feedback?

astojilj:
About upstream: initiated the CL discussion here: 
https://codereview.chromium.org/1746013002
I'll submit the bug and handle the Linux case there too. It is a bug - don't see it's a problem to get people interested to get it upstream.

ui/views/controls/native/native_view_host.cc
```diff
@@ -127,6 +127,23 @@ void NativeViewHost::VisibilityChanged(View* starting_from, bool is_visible) {
   Layout();
 }
 
+void NativeViewHost::OnSoftVisibilityChanged(bool visible) {
+  if (!native_view_ || !native_wrapper_.get())
```
pozdnyakov: nit: .get()is not needed

ui/views/controls/native/native_view_host.cc
```diff
@@ -127,6 +127,23 @@ void NativeViewHost::VisibilityChanged(View* starting_from, bool is_visible) {
...
+    return;
+
+  const bool os_visible = !GetVisibleBounds().IsEmpty();
```
pozdnyakov: if (GetVisibleBounds().IsEmpty())
  return;

ui/views/view.cc
```diff
@@ -1840,6 +1840,12 @@ void View::PropagateNativeThemeChanged(const ui::NativeTheme* theme) {
   OnNativeThemeChanged(theme);
 }
 
+void View::PropagateSoftVisibilityChanged(bool visible) {
+  for (int i = 0, count = child_count(); i < count; ++i)
```
pozdnyakov: do we really win anything caching 'count'? for (int i = 0; i < child_count(); ++i) ?

ui/views/view.h
```diff
@@ -1073,6 +1073,8 @@ class VIEWS_EXPORT View : public ui::LayerDelegate,
   // FocusManager manages this view.
   virtual void NativeViewHierarchyChanged();
 
+  virtual void OnSoftVisibilityChanged(bool visible) {}
```
pozdnyakov: the other methods are not defined in .h, the new method should take after them

ui/views/widget/widget.cc
```diff
@@ -1086,6 +1086,12 @@ void Widget::OnNativeWidgetVisibilityChanged(bool visible) {
     root->layer()->SetVisible(visible);
 }
 
+void Widget::OnSoftVisibilityChanged(bool visible) {
+  View* root = GetRootView();
+  if (root)
```
pozdnyakov: if (View* root = GetRootView())
  root->

ui/views/controls/native/native_view_host.cc
astojilj: I would keep it - as it is on 3 other places like that if ... native_wrapper_.get()

ui/views/controls/native/native_view_host.cc
astojilj: Done.

ui/views/view.cc
astojilj: Don't know. As on 16 other places in that file (where I copy pasted the code) it is like this, suggest to keep the style.

ui/views/view.h
astojilj: There are 9 empty {} implementations in view.h

ui/views/widget/widget.cc
astojilj: Done.

astojilj:
@pozdnyakov Thanks. Updated patch; 2 issues fixed, didn't act on some of the nits.

pozdnyakov:
lgtm

PR #325, opened 2016-03-07
[Lite] Set aapt to do compressing on all the apk contents.

heke123:
Apk is actually a common zip file. Since Lite-17, the default
behavior of generating apk has been changed as not compressing
.dat .pak and .bin files. So,the final apk size is much bigger.

Take Lite for example, xwalk.pak file is ~800K, if we compress
it when generating apk, the final apk size is 500K smaller.

### DIFF
--- a/build/java_apk.gypi
+++ b/build/java_apk.gypi
@@ -94,7 +94,7 @@
     'is_test_apk%': 0,
     # Allow icu data, v8 snapshots, and pak files to be loaded directly from the .apk.
     # Note: These are actually suffix matches, not necessarily extensions.
-    'extensions_to_not_compress%': '.dat,.bin,.pak',
+    'extensions_to_not_compress%': '',
     'resource_input_paths': [],
     'intermediate_dir': '<(PRODUCT_DIR)/<(_target_name)',
     'asset_location%': '<(intermediate_dir)/assets',


heke123:
I didn't check the crosswalk-16, I found in crosswalk-15, all the files will be zipped, but in crosswalk-17, .dat .pak .bin files are skipped when zipping.
In crosswalk master, the ICU feature is not cut off, the icudtl.dat is more than 6M and not compressed since crosswalk-17(or 16). It's one of the reason that the binary size has a sharp increase. Of course the content_shell's size also increased because it shares the same apk packaging gypi file.

heke123:
@halton  @PeterWangIntel  Please help review. thanks.

PeterWangIntel:
LGTM
Thank you very much.

heke123:
Sorry... Just found it seems only meaningful with our test apks such as xwalkcoreshell.apk and xwalkruntimeEmbedded.apk.  The final aar and apk size are not effected !

halton:
@heke123 will the .pak be depressed and loaded correctly?

heke123:
@halton  Is it meaningful if it only works for XWalkCoreShell.apk?  If not, I'll drop this PR.  Thanks.

heke123:
Yes, the .pak can be loaded correctly.

halton:
Does not core_library include xpk file? If yest, I think this reduction is very huge and value to affect the packaged app as well.

heke123:
Java_apk.gypi is an apk-packing tool which is used by XWalkCoreShell, Content_shell, XWalkEmbeddedShell etc.
1) If user wants to embed xwalk_core_library (or AAR) in their project, they use their IDE to generate apk(Eclipse or AndroidStudio or Android SDK Command Line).  The apk will NOT be shrinked.
2) If user chooses make_apk.py or App_tool to generate apk, (make_apk.py use Android SDK Command Line). It also have nothing to do with java_apk.gypi. The apk will NOT be shrinked either.

halton:
LGTM

halton:
Please continue investigate whether the pak resouce files are compressed or not.

heke123:
No problem, I'll continue to investigate on this. Thanks very much for your reminder :)

halton:
@heke123 please remember to bump the DEPS file

PR #326, opened 2016-03-07
Windows: PageVisibility API when any other window goes fullscreen

astojilj:
BUG=XWALK-5503

Some OSs (e.g. Windows and Linux) don't mark windows as hidden on screen
lock or when other opaque windows fully cover them.

The patch here targets Windows and window from any process going fullscreen.
For this, AppBar notifications are used.
Note that this covers e.g. VLC, MS Paint or browsers going fullscreen. It
doesn't cover screensaver or when application windows are maximized.

In the case when current window goes fullscreen, appbar notificaion is also
received. This case is handled, too.


astojilj:
@darktears, @pozdnyakov 
Could you please check this addition. Thanks

PR #327, opened 2016-03-10
[Android] Fix the lint warnings for Media classes.

wuhengzhi:
Adapt to:
  https://codereview.chromium.org/1506013009
  https://codereview.chromium.org/1588503004
Some interfaces for media were added from API level 16, we need to check
it at runtime otherwise app will be crashed when it runs on ICS.
Add the checks for API level, if API level is less than JELLY_BEAN,
return default value.

BUG=XWALK-6494

### DIFF
--- a/media/base/android/java/src/org/chromium/media/MediaPlayerBridge.java
+++ b/media/base/android/java/src/org/chromium/media/MediaPlayerBridge.java
@@ -121,6 +121,7 @@ public class MediaPlayerBridge {
     }
 
     private boolean hasTrack(int trackType) {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN) return false;
         try {
             TrackInfo trackInfo[] = getLocalPlayer().getTrackInfo();
 
--- a/media/base/android/java/src/org/chromium/media/MediaCodecUtil.java
+++ b/media/base/android/java/src/org/chromium/media/MediaCodecUtil.java
@@ -77,6 +77,11 @@ class MediaCodecUtil {
     @SuppressWarnings("deprecation")
     @CalledByNative
     private static CodecInfo[] getCodecsInfo() {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN) {
+            // It's better to return empty list instead of null pointer.
+            CodecInfo[] emptyInfo = {};
+            return emptyInfo;
+        }
         // Return the first (highest-priority) codec for each MIME type.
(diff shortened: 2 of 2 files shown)


rakuco:
What is the impact of essentially always returning `false` or empty lists on ICS? It looks like we're basically making some functionality not work as expected.

media/base/android/java/src/org/chromium/media/MediaPlayerBridge.java
```diff
@@ -122,17 +122,19 @@ protected boolean hasAudio() {
 
     private boolean hasTrack(int trackType) {
         try {
-            TrackInfo trackInfo[] = getLocalPlayer().getTrackInfo();
```
lincsoon: I think the code below with less changes would be better.

```
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) return false
```

lincsoon:
@rakuco The methods being modified this time are like `canDecode` or `getSupportedCode`, but not something must expect clear result. Even the device is above 4.0, it also could return false. And each of these methods give default return value too.

lincsoon:
@wuhengzhi Based on your changes, I can see that if the device is 4.0, the method almost do nothing. For these situations, it's better to check the version and return default value at first to reduce the code change, also reduce the possibility of further conflict.

wuhengzhi:
@lincsoon Yes, thanks.

wuhengzhi:
@rakuco @lincsoon PTAL, thanks.

lincsoon:
LGTM

rakuco:
> @rakuco The methods being modified this time are like canDecode or getSupportedCode, but not something must expect clear result. Even the device is above 4.0, it also could return false. And each of these methods give default return value too.

My question was more along these lines: did the code that uses these new methods receive a different, working value before M49? In other words, does this new code result in things that used to work before not working anymore?

wuhengzhi:
@rakuco All of the checks just for the API level is less than Jelly Bean.
Before M49, if app runs on Android system which version is less than 4.1, it will be crashed without these checks.

> In other words, does this new code result in things that used to work before not working anymore?

So these checks won't impact the result which API level is equal or greater than JB.

rakuco:
> @rakuco All of the checks just for the API level is less than Jelly Bean.

I know, I'm only talking about the ICS case which is the reason we need to change this code in the first place.

> Before M49, if app runs on Android system which version is less than 4.1, it will be crashed without these checks.

What I've been saying from the beginning: this code as-is _does not exist_ before M49. There were no lint issues in the media code before M49. So what I'm asking is if, compared to the previous version of the code that did not use API not present in ICS, this patch reduces functionality on ICS.

PR #334, opened 2016-03-30
WebCL: Import OpenCL headers into chromium-crosswalk.

rakuco:
With M49, the WebCL code is always being built, so we need the OpenCL
headers here for pure chromium-crosswalk content_shell builds to work,
since they do not check out Crosswalk and pull its dependencies.

BUG=XWALK-6661

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -338,7 +338,6 @@ vs-chromium-project.txt
 /third_party/junit/src
 /third_party/kasko
 /third_party/khronos_glcts
-/third_party/khronos/CL
 /third_party/leveldatabase/src
 /third_party/leveldb
 /third_party/libc++-static/libc++.a
--- a/third_party/khronos/CL/cl.h
+++ b/third_party/khronos/CL/cl.h
@@ -1,0 +1,1214 @@
+/*******************************************************************************
+ * Copyright (c) 2008 - 2012 The Khronos Group Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and/or associated documentation files (the
+ * "Materials"), to deal in the Materials without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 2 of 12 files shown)


rakuco:
@shaoboyan @hujiajie

shaoboyan:
LGTM, thx.

hujiajie:
Hi @rakuco, since I'm not quite familiar with the buildbot and gclient, can you explain the following questions:
1. Will the coding style be checked in the future after they get merged? If the answer is no, then it LGTM.
2. I suppose crosswalk.git will be updated with two commits in order later: a) roll chromium-crosswalk (say the commit id in crosswalk.git is abc1234); b) remove the opencl header entry in DEPS.xwalk (def5678). Then if some one is trying to sync crosswalk@abc1234, will these headers conflict with the residual entry in DEPS.xwalk?
3. Will .git in src/third_party/khronos/CL be removed automatically after commit def5678?

Thanks

rakuco:
> 1. Will the coding style be checked in the future after they get merged? If the answer is no, then it LGTM.

The coding style is checked whenever a patch touches a certain file. With our current infrastructure, it means the style checker will complain if we ever change those headers again. This is less than ideal, so I've decided to work on XWALK-1853 and replace the current style checker with {{PRESUBMIT.py}}.

> 2. I suppose crosswalk.git will be updated with two commits in order later: a) roll chromium-crosswalk (say the commit id in crosswalk.git is abc1234); b) remove the opencl header entry in DEPS.xwalk (def5678). Then if some one is trying to sync crosswalk@abc1234, will these headers conflict with the residual entry in DEPS.xwalk?
> 3. Will .git in src/third_party/khronos/CL be removed automatically after commit def5678?

There will be a single commit in Crosswalk doing both a) and b). depot_tools should take care of everything automatically (it will first remove the `src/third_party/khronos/CL` git checkout and then its contents will be readded when chromium-crosswalk is updated).

hujiajie:
LGTM. Thanks

PR #335, opened 2016-04-01
[Windows] Fix build.

darktears:
Subclass was exported but not the base class. It triggers a warning
which stops the build.

Fixed upstream by CL https://codereview.chromium.org/1637273003


darktears:
@rakuco PTAL

rakuco:
le lgtm :fr:

rakuco:
The style errors are false positives that have been fixed since then by running `PRESUBMIT.py` instead of our homegrown `lint.py` (XWALK-1853). Merging.

PR #338, opened 2016-04-06
[WebCL] Enable GN build.

hujiajie:
This commit enables GN build of the WebCL module. Files for custom
bindings are the only ones that need to be added to GN
configuration. All other files of the WebCL module are imported from GYP
configurations automatically. The patch has been tested with a vanilla
Chromium + WebCL build.


hujiajie:
@shaoboyan @PeterWangIntel @rakuco

shaoboyan:
LGTM.

PR #339, opened 2016-04-06
[WebCL] Let WebCLEvent.prototype.getProfilingInfo() return 64-bit pro…

hujiajie:
…filing info.

The profiling info of an event is a 64-bit value that describes the
current device time counter in nanoseconds
(https://www.khronos.org/registry/cl/sdk/1.1/docs/man/xhtml/clGetEventProfilingInfo.html),
but it was casted as a 32-bit integer in previous implementation, and
thus made the return value almost useless. I believe this is a typo in
the WebCL specification, and this commit fixes the problem.

### DIFF
--- a/third_party/WebKit/Source/modules/webcl/WebCLEvent.cpp
+++ b/third_party/WebKit/Source/modules/webcl/WebCLEvent.cpp
@@ -84,18 +84,20 @@ int WebCLEvent::getStatus()
     return CL_INVALID_VALUE;
 }
 
-unsigned WebCLEvent::getProfilingInfo(int paramName, ExceptionState& es)
+ScriptValue WebCLEvent::getProfilingInfo(ScriptState* scriptState, unsigned paramName, ExceptionState& es)
 {
+    v8::Isolate* isolate = scriptState->isolate();
+
     if (isReleased()) {
         es.throwWebCLException(WebCLException::INVALID_EVENT, WebCLException::invalidEventMessage);
-        return 0;
+        return ScriptValue(scriptState, v8::Null(isolate));
     }
 
     int status = getStatus();
     unsigned properties = m_commandQueue ? m_commandQueue->getProperties() : 0;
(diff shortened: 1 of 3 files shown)


hujiajie:
@shaoboyan PTAL

shaoboyan:
LGTM

rakuco:
Is there a test or tracking bug about this? It's not clear to me where the value was being cast to a 32-bit value before, as it was already using `unsigned long long`.

hujiajie:
I've just create [XWALK-6075](https://crosswalk-project.org/jira/browse/XWALK-6705) to track this issue.
Previously, the return type of `WebCLEvent::getProfilingInfo()` is `unsigned` to fit the IDL interface, so `static_cast<unsigned long long>()` does not help at last.
The commit message has been updated a bit accordingly.

rakuco:
> Previously, the return type of WebCLEvent::getProfilingInfo() is unsigned to fit the IDL interface, so static_cast<unsigned long long>() does not help at last.

I see, thanks for the explanation. Doesn't it work if you change `WebCLEvent::getProfilingInfo()`'s return type to `unsigned long long` instead of `ScriptValue`?

hujiajie:
The direct reason is that the method is defined as `any getProfilingInfo(CLenum name)` in [WebCL specification](https://www.khronos.org/registry/webcl/specs/1.0.0/), and `any` is represented with `ScriptValue` in Blink.

To further understand the rationality, by design there's no guarantee that profiling info is always a 64-bit unsigned integer. This can be inferred from the corresponding C declaration in [OpenCL specification](https://www.khronos.org/registry/cl/sdk/1.1/docs/man/xhtml/):

```
cl_int clGetEventProfilingInfo(	cl_event event,
    cl_profiling_info param_name,
    size_t param_value_size,
    void *param_value, // `void *` instead of `unsigned long long *`
    size_t *param_value_size_ret)
```

It may happen that future specification will allow programmers to query other profiling info which is not a 64-bit integer, so I prefer to use a generic type here.

rakuco:
Makes sense, thanks for the explanation!

PR #341, opened 2016-04-07
Enable H264 in WebRTC

js0701:
Cherry pick 49485b38 to enable Android OpenMax decoder for H264
Update WebRTC and libjingle repo with minimum version which begins to support H264 codec behind a rtc_use_h264 build flag. The flag is off by default, and need a patch in crosswalk gyp_xwalk to turn it on.
BUG=XWALK-2310, XWALK-6627


js0701:
@halton @rakuco PTAL

hthetiot:
https://bugs.chromium.org/p/chromium/issues/detail?id=719023

rakuco:
Closing, Crosswalk is unmaintained and we won't be making any new releases.

PR #342, opened 2016-04-13
Fix calling of AudioOutputStream::AudioSourceCallback API

pozdnyakov:
Before this change calling of AudioOutputStream::AudioSourceCallback
method without stream position ended up in an empty implementation.

BUG=XWALK-6703


pozdnyakov:
@darktears  @rakuco PTAL

rakuco:
rs=me

nerdmed:
@rakuco would be great to see this rolled out so the master very soon, i will double check the fix and report if it fixes the issue

PR #344, opened 2016-04-21
Rebase WebCL on M50.

shaoboyan:
Rebase  WebCL on M50

   Adapted to the following changes:
    WebCL: Adding missing virtual final destructors for corrseponding
    classes to avoid possible destruction unexpected behaviour.
    WebCL: Change WebCLCallback parent class from RefCounted<T> to
    GarbageCollectedFinalized<T> for idl generation rule changed. Now the
    upstream generate idl interface with callback attribute with oilpan
    support.
    WebCL: Due to WebCLCallback is oilpan supported, modified RefPtr to
    Persistent in corresponding non-garbage-collected class to elimate
    compile error.


shaoboyan:
@rakuco @hujiajie  PTAL

hujiajie:
@shaoboyan it seems that you didn't change `Vector<RefPtr<...>>` to `HeapVector<Member<...>>` and its friends. Please double check that.

shaoboyan:
@hujiajie  done.

hujiajie:
There's no need to make the destructors virtual. I think make the classes final should suppress most of the warnings.

PR #345, opened 2016-04-22
WebCL Rebase M50

shaoboyan:
Rebase WebCL on M50

Adapted to the following changes:
WebCL: Adding missing virtual final destructors for corrseponding
classes to avoid possible destruction unexpected behaviour.
WebCL: Change WebCLCallback parent class from RefCounted to
GarbageCollectedFinalized for idl generation rule changed. Now the
upstream generate idl interface with callback attribute with oilpan
support.
WebCL: Due to WebCLCallback is oilpan supported, modified RefPtr to
Persistent in corresponding non-garbage-collected class to elimate
compile error.


shaoboyan:
@hujiajie @rakuco  PTAL

third_party/WebKit/Source/bindings/modules/v8/custom/custom.gni
```diff
@@ -9,5 +9,7 @@ bindings_modules_v8_custom_files =
                     "V8DeviceMotionEventCustom.cpp",
                     "V8ExtendableMessageEventCustom.cpp",
                     "V8ServiceWorkerMessageEventCustom.cpp",
+                    "V8WebCLCommandQueueCustom.cpp",
+                    "V8WebCLProgramCustom.cpp",
```
hujiajie: This should belong to previous commit.

hujiajie:
1. There's no need to replace `adoptRef()` with explicit `Member<>()` or `Persistent<>()`, just use raw pointer there.
2. Please make sure either Oilpan style or Oilpan-transition style (FooWillBeBar) is used, but not both.

PR #348, opened 2016-04-27
[Android] Expose headers parameter to be used by MediaPlayer

fujunwei:
Give the host application a chance to take over the control when a new
media is about to be played in the current XWalkViewInternal.

BUG=XWALK-6770


fujunwei:
@rakuco @lincsoon @axinging @wuhengzhi PTAL, thanks.

media/base/android/java/src/org/chromium/media/MediaPlayerBridge.java
```diff
@@ -28,17 +28,20 @@
...
 @JNINamespace("media")
 public class MediaPlayerBridge {
-
+    /**
+     * Give the host application a chance to take over the control.
+     */
     public static class ResourceLoadingFilter {
         public boolean shouldOverrideResourceLoading(
-                MediaPlayer mediaPlayer, Context context, Uri uri) {
+                MediaPlayer mediaPlayer, Context context, Uri uri, Map<String, String> headers) {
```
lincsoon: Is `headers` a mandatory parameter? Why don't let the old constructor and the new one coexist?

media/base/android/java/src/org/chromium/media/MediaPlayerBridge.java
```diff
@@ -203,9 +206,9 @@ protected boolean setDataSource(
             headersMap.put("allow-cross-domain-redirect", "false");
         }
         try {
-            if (sResourceLoadFilter != null &&
-                    sResourceLoadFilter.shouldOverrideResourceLoading(
-                            getLocalPlayer(), context, uri)) {
+            if (sResourceLoadFilter != null
+                    && sResourceLoadFilter.shouldOverrideResourceLoading(
```
lincsoon: These two lines don't need to be modified.

media/base/android/java/src/org/chromium/media/MediaPlayerBridge.java
```diff
@@ -203,9 +206,9 @@ protected boolean setDataSource(
...
+                               getLocalPlayer(), context, uri, headersMap)) {
```
lincsoon: Extra spaces at the beginning of the line

media/base/android/java/src/org/chromium/media/MediaPlayerBridge.java
lincsoon: Is `headers` a mandatory parameter? Why don't let the method has two interface. If upstream use the old interface in the future, there will be a conflict

wuhengzhi: ```ResourceLoadingFilter``` was added by Crosswalk to resolve media loading issues.

PR #351, opened 2016-05-04
[ICS Support] Remove ffmpeg from DEPS

lincsoon:
The original ffmpeg uses the system call "posix_memalign" that
doesn't exist on Android 4.0. To use modified ffmpeg code,remove
ffmpeg from chromium's DEPS and fetch forked repo through
crosswalk's DEPS.

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -79,8 +79,6 @@ deps = {
     (Var("chromium_git")) + '/external/colorama.git@799604a1041e9b3bc5d2789ecbd7e8db2e18e6b8',
   'src/third_party/dom_distiller_js/dist':
     (Var("chromium_git")) + '/external/github.com/chromium/dom-distiller-dist.git@e21fe06cb71327ec62431f823e783d7b02f97b26',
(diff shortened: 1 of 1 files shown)


lincsoon:
@rakuco Please help to merge

rakuco:
I don't get why this is necessary. Changing `DEPS.xwalk` in Crosswalk just like we do for v8-crosswalk should be enough.

lincsoon:
I was going to add corresponding dependency to `DEPS.xwalk` too. But I'm not sure the mechanism behind here. What if one repository exists in both of chromium's DEPS and crosswalk's DEPS, and the commit is different? Does it make any conflict?

rakuco:
> What if one repository exists in both of chromium's DEPS and crosswalk's DEPS, and the commits are different? Does it make any conflict?

No, otherwise v8-crosswalk would never have worked at all. The definitions in `DEPS.xwalk` always override the ones in chromium-crosswalk's `DEPS`.

PR #353, opened 2016-05-17
[Android] Support third part media player on Crosswalk

fujunwei:
A requirement from an important customer who want to forward web resources with proxy
on Crosswalk, but android system MediaPlayer can't set a proxy with a standard API.
The ExoMediaPlayer is playing videos and music is a popular activity on Android devices,
and it can be configured with proxy.
https://developer.android.com/guide/topics/media/exoplayer.html

BUG=XWALK-6770


fujunwei:
@rakuco @lincsoon @axinging @wuhengzhi PTAL.

baleboy:
Crosswalk 20 branch date is May 27th, please prioritize the review if you want this in.

media/base/android/java/src/org/chromium/media/MediaPlayerBridge.java
```diff
@@ -78,22 +84,22 @@ protected void destroy() {
         mNativeMediaPlayerBridge = 0;
     }
 
-    protected MediaPlayer getLocalPlayer() {
-        if (mPlayer == null) {
-            mPlayer = new MediaPlayer();
+    protected ExMediaPlayer getExMediaPlayer() {
```
axinging: No need to change name, getLocalPlayer will also be ok. 
Is there an easy way to switch from MediaPlayer to ExoMediaPlayer, , co-exist?

wuhengzhi: Is it better to create `ExMediaPlayerBridge.java` to reduce the rebase workload?

wuhengzhi:
How to set the proxy for player? I can not find the interface.

media/base/android/java/src/org/chromium/media/MediaPlayerBridge.java
fujunwei: How to use it in MediaPlayerBridge?

fujunwei:
The API will to be exposed in another PR in crosswalk repo.

media/base/android/java/src/org/chromium/media/MediaPlayerBridge.java
fujunwei: I agree to keep the original name to reduce rebase workload.

fujunwei:
The PR of exposing api to developer is https://github.com/crosswalk-project/crosswalk/pull/3736

fujunwei:
@rakuco @lincsoon @axinging @wuhengzhi PTAL

media/base/android/java/src/org/chromium/media/ExMediaPlayer.java
```diff
@@ -0,0 +1,123 @@
+// Copyright 2013 The Chromium Authors. All rights reserved.
```
wuhengzhi: 2016, add Intel's copyright?

fujunwei:
Thanks, done.

media/base/android/java/src/org/chromium/media/MediaPlayerBridge.java
```diff
@@ -202,9 +208,9 @@ protected boolean setDataSource(
             headersMap.put("allow-cross-domain-redirect", "false");
         }
         try {
-            if (sResourceLoadFilter != null &&
-                    sResourceLoadFilter.shouldOverrideResourceLoading(
-                            getLocalPlayer(), context, uri)) {
+            if (sResourceLoadFilter != null
+                    && sResourceLoadFilter.shouldOverrideResourceLoading(
+                               getLocalPlayer(), context, uri)) {
```
lincsoon: Looks like you changed nothing but still left a diff record here.

fujunwei: It's for passing style checks.

fujunwei:
@rakuco @halton @lincsoon @axinging @wuhengzhi Do you have any updates?

media/base/android/java/src/org/chromium/media/ExternalMediaPlayer.java
```diff
@@ -0,0 +1,123 @@
...
+*/
+public class ExternalMediaPlayer {
+    private static final String TAG = "ExternalMediaPlayer";
+
+    private MediaPlayer mPlayer;
+
+    protected MediaPlayer getLocalPlayer() {
+        if (mPlayer == null) {
+            mPlayer = new MediaPlayer();
+            Log.d(TAG, "Create a Android System Media Player.");
```
wuhengzhi: `a` -> `an`.

media/base/android/java/src/org/chromium/media/MediaPlayerBridge.java
```diff
@@ -35,12 +35,18 @@
...
 @JNINamespace("media")
 public class MediaPlayerBridge {
-
+    /**
+    * Give the host application a chance to take over MeidaPlayer.
+    */
     public static class ResourceLoadingFilter {
         public boolean shouldOverrideResourceLoading(
-                MediaPlayer mediaPlayer, Context context, Uri uri) {
+                ExternalMediaPlayer mediaPlayer, Context context, Uri uri) {
```
axinging: Sorry for the late response.
Is there any switch to fallback to MediaPlayer? It seems API changed. I prefer to add a new API for this.

fujunwei: We don't know which MediaPlayer is using, Maybe a flag need to be added, so it's more complicated.

axinging:
LGTM

darktears:
LGTM.

@rakuco ?

rakuco:
Not yet, IMO.

* The questions that must always be asked on every single chromium-crosswalk pull request and which have not been addressed here: have you tried to upstream this in any form? If not, why not? What's the maintenance plan here, are we expected to carry this patch forever?
* "I want to integrate this patch because my unnamed, important customer wants it" is not a valid reason to have a patch merged into a public, open-source project. You need to come up with a better explanation of why we want this.

Code-wise: isn't it possible to add a delegate that returns a subclass of `MediaPlayer` so you don't need to define `ExternalMediaPlayer` at all? Something like making `getLocalPlayer()` return either a new `MediaPlayer` or a subclass if one is registered in Crosswalk. This way the change to chromium-crosswalk is much smaller (and easier to get accepted upstream) and you make `XWalkMediaPlayer` inherit from `MediaPlayer` and let your users change its behavior as they see fit.

fujunwei:
> The questions that must always be asked on every single chromium-crosswalk pull request and which have not been addressed here: have you tried to upstream this in any form? If not, why not? What's the maintenance plan here, are we expected to carry this patch forever?

I will communicate with upstream to figure out whether it's acceptable feature.

> "I want to integrate this patch because my unnamed, important customer wants it" is not a valid reason to have a patch merged into a public, open-source project. You need to come up with a better explanation of why we want this.

Thanks Sunlin's explanation: This feature is requested by one of our customers. They came to us because they couldn't realize their idea with existing solutions. First, there's no proxy configuration for Android WebView, the developer can redirect the reqeust for specific url but there's nothing they can do about video streaming, which is just what the customer wants. Second, although chromium supports proxy settings, but it's not available on Android.

> isn't it possible to add a delegate that returns a subclass of MediaPlayer so you don't need to define ExternalMediaPlayer at all? 

The approach has a demerit that an invalid MediaPlayer is created by default, but the third party MediaPlayer don't need it usually such as Google ExoPlayer. The refined codes is also committed, PTAL.

SonicACCEL:
Is there a status update on this? 

https://stackoverflow.com/questions/49394213/using-exoplayer-for-video-playback-in-a-crosswalk-project-android-application

PR #354, opened 2016-05-17
[Lite] remove duplicate defined resources

tingshao:
GYP complains when setting use_minimum_resource=0. Two resources are
duplicate defined: IDR_MENU_CHECK_CHECKED and IDR_MENU_DROPARROW.
That's because they have already been defined no matter use_minimum_resource
is switched on/off.

The definition inside the condition block for use_minimum_resource is removed. 
@PeterWangIntel @heke123 @halton


heke123:
LGTM

PR #355, opened 2016-05-23
[WebCL] Include what you use.

hujiajie:
The patch set aims to fix most (if not all) style issues to follow the principle of including what you use. The implementation of some non-trivial methods / ctors / dtors are moved from .h to .cpp so that the principle can be well followed. A self-contained header should include the necessary headers so that we can simply use it in other files without worrying about its dependencies. On the other hand, a header file should only include what is really needed and use forward declarations for external symbols if possible. This will not only help in faster compilation and fewer recompilation, but also avoid circular header inclusion which can actually cause compilation errors that are hard to diagnose in some cases. Also, if a header file is needed, then it should be included *explicitly*, instead of depending on the fact that it is *implicitly* included by some other random files which are subject to change in the future.


hujiajie:
@shaoboyan PTAL

hujiajie:
ping @shaoboyan

shaoboyan:
LGTM, thx

hujiajie:
@rakuco PTAL. If there's no problem, I'll open a new PR for the next branch to eliminate similar issues in M51 rebase.

rakuco:
M51 hit master last week, so there's no need to worry about next any longer. You do need, however, to reupload your patch after rebasing it on top of the current master branch.

rakuco:
master is now on M52. Are there any plans to continue work on this PR or can it be discarded?

hujiajie:
This PR is no longer valid. It requires a new round of audit to find out all the headers needed in each file. I'll open a new PR when that's done, and maybe split the work into smaller pieces to make it easier to land in a gradual way.

PR #356, opened 2016-06-06
Remove .gclient and .gclient_entries.

rakuco:
Those two files were erroneously added in commit bbc2a01 ("[Temp] Do not
enable generate_multidex_config by default on Android").

Related BUG=XWALK-6547
BUG=XWALK-6625
BUG=XWALK-6961


mrunalk:
LGTM

PR #363, opened 2016-06-28
[Android] Support the feature of proxy in Crosswalk

fujunwei:
Crosswalk for android will update proxy when a PROXY_CHANGE intent is received,
but it’s need system permission to send the broadcast, there is also a workaround with
reflecting the incompatible api, but it’s too complex to maintain those codes, so
implement the new API in XWalkView to set proxy for images/text resources.

BUG=XWALK-6769


fujunwei:
@lincsoon @xuxin @wuhengzhi PTAL.

lincsoon:
LGTM @rakuco

rakuco:
The questions that must always be asked on every single chromium-crosswalk pull request and which have not been addressed here: have you tried to upstream this in any form? If not, why not? What's the maintenance plan here, are we expected to carry this patch forever?

rakuco:
cc @darktears

lincsoon:
> have you tried to upstream this in any form? If not, why not?

This feature is requested by one of our customers. They came to us because they couldn't realize their idea with existing solutions. First, there's no proxy configuration for Android WebView, the developer can redirect the reqeust for specific url but there's nothing they can do about video streaming, which is just what the customer wants. Second, although chromium supports proxy settings, but it's not available on Android.

> What's the maintenance plan here, are we expected to carry this patch forever?

As you can see, we just added some new methods in a class instead of modifying original logic. As long as this class is at work, we want to carry on this feature as a good differentiation from default WebView.

PR #369, opened 2016-07-19
HLS metadata parsing

srikumarb:
Android Framework does not parse metadata for HLS playlists
hence added the required support in chromium MediaResourceGetter


rakuco:
@srikumarb: are you still interested in working on this? It's not clear if this has been submitted upstream, if there's an associated ticket in our bug tracker or why exactly we should have this in chromium-crosswalk.

srikumarb:
I am planning to promote these changed to crosswalk/chromium-crosswalk. I have created a fork and made the changes there.

PR #370, opened 2016-07-25
Add gn build support for crosswalk-linux

heke123:
BUG=XWALK-7180


heke123:
The modification in BUILD.gn seems messy,  actually all the modifications is like below:
--------------start  -------------
declare_args() {
  build_xwalk = ""
}
if (build_xwalk) {
  group("xwalk") {
    testonly = true
    deps = [
      "xwalk:xwalk",
    ]
 }
}
if (!build_xwalk) {
   .... the original contents...
}
--------------end --------------

I use the build_xwalk variable to controll whether to build crosswalk  or  upstream targets like  content_shell.
I run the "gn format" command on BUILD.gn because our build server has the format check. that's why it looks messy.

heke123:
@rakuco @halton @PeterWangIntel Please have a look, thanks.

PeterWangIntel:
LGTM， as our offline discussion. Maybe, the splited patch is eaier to be review.

PeterWangIntel:
LGTM

rakuco:
A less intrusive approach is to just add a `xwalk` directory to `build/secondary` with one or more empty targets with the same name as the ones you want in` src/xwalk/BUILD.gn` itself. This way, if only chromium-crosswalk is checked out and `src/xwalk` is not present, the targets in `build/secondary` will be used, otherwise the real versions in `src/xwalk` are used instead.

build/config/features.gni
```diff
@@ -172,8 +172,9 @@ enable_chromevox_next = false
 # Use brlapi from brltty for braille display support.
 use_brlapi = is_chromeos
 
-# Enable WebVR support by default on Android
-# Still requires command line flag to access API
-enable_webvr = is_android
+declare_args() {
```
rakuco: You could have at least reused the same `declare_args` block that's in the beginning of the file.

Anyway, this is exactly the kind of thing you should attempt to send upstream: https://codereview.chromium.org/2198003002/

rakuco:
> Testing patch series with heke123/chromium-crosswalk@78e49e1 as its head.

See my previous comment about the `build/secondary` approach. Changing `BUILD.gn` like that is a complete no-go and totally unmaintainable across rebases.

build_overrides/v8.gni
```diff
@@ -12,7 +12,10 @@ enable_correct_v8_arch = false
 # TODO: Set v8_use_external_startup_data to true on Windows and iOS to match
 # GYP.
 # iOS - http://crbug.com/545676
-v8_use_external_startup_data = !is_ios
+declare_args() {
```
rakuco: This could be replaced by a backport of [this upstream commit](https://crrev.com/0fffeb2adaa3c284b760922c1aecce1516b998ce).

rakuco:
> See my previous comment about the build/secondary approach. Changing BUILD.gn like that is a complete no-go and totally unmaintainable across rebases.

And another, simpler approach, is to just change `root_extra_deps` in Crosswalk and not do anything in `src/BUILD.gn` itself.

PR #371, opened 2016-08-08
Change locale path from "locales" to "locales/xwalk"

heke123:
Both Crosswalk and Chrome they place the en-US.pak into folder
"locales" in build of Linux and windows. Potientially there is
conflicts, so it's better to set Crosswalk's locale-path as
"locales/xwalk".

### DIFF
--- a/ui/base/ui_base_paths.cc
+++ b/ui/base/ui_base_paths.cc
@@ -37,6 +37,7 @@ bool PathProvider(int key, base::FilePath* result) {
         return false;
 #else
       cur = cur.Append(FILE_PATH_LITERAL("locales"));
+      cur = cur.Append(FILE_PATH_LITERAL("xwalk"));
 #endif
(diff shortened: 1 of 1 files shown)


heke123:
@rakuco   Thanks.

heke123:
Call stack of get locale path:
ResourceBundle::LoadLocaleResources
    ResourceBundle::GetLocaleFilePath()
         PathService::Get(ui::DIR_LOCALES, &locale_file_path);

rakuco:
Hardcoding the paths like that will break any non-Crosswalk (e.g content shell) builds. What you need is to subclass `ui::ResourceBundle::Delegate` and implement its `GetPathForLocalePack()` method in Crosswalk. There's no need to change anything here.

You can see an example of what I'm talking about in `src/chromecast/common/cast_resource_delegate.{cc,h}`.

PR #373, opened 2016-08-12
Update chromium to 52.0.2743.116

msisov:


mrunalk:
You will need to force push this update directly to master as these updates are not totally additive. Also before you force push your changes to master you need to take a backup existing master. Ofcourse please check-in with @rakuco before you force push.

PR #374, opened 2016-08-16
[Printing] Disable print preview to enable basic printing

mrunalk:
This is related to https://github.com/crosswalk-project/crosswalk/pull/3851
to enable window.print() functionality.

BUG=XWALK-5680

### DIFF
--- a/build/common.gypi
+++ b/build/common.gypi
@@ -551,7 +551,7 @@
 
       # Enable printing with print preview. It does not imply
       # enable_basic_printing. It's possible to build Chrome with preview only.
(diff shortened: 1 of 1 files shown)


rakuco:
Why not change the default in Crosswalk's `build/common.gypi`? It'd also be good to explain why the print code is broken when this flag is on.

mrunalk:
Ahh I didn't know crosswalk has it's own build/common.gypi too. Thanks. I will close this PR and add this change to the other PR with additional notes as to why we need this change.

PR #375, opened 2016-08-17
Add GN build support for crosswalk-linux

heke123:
As Chromium will deprecate GYP build system since M54, we need to enable
GN build support for Crosswalk as soon as possible.
Set xwalk as "extra_deps" to Chromium in xwalk/build/common.gni:
`root_extra_deps = [ "//xwalk:xwalk_builder" ]`

### DIFF
--- a/.gn
+++ b/.gn
@@ -311,4 +311,5 @@ exec_script_whitelist = [
   "//tools/gn/setup.cc",
   "//ui/accessibility/BUILD.gn",
   "//ui/views/BUILD.gn",
+  "//xwalk/build/version.gni",
 ]


heke123:
@rakuco

rakuco:
The changes look fine to me, but they need to be committed differently: the `enable_webvr` part should be a separate commit backported from upstream (use `git cherry-pick`, edit the commit message to mention it is a backport like [this one](https://github.com/crosswalk-project/chromium-crosswalk/commit/701504902f6ee7676d020ab88ff69982cb7426b4) and make sure it does not contain the "Cr-Commit-Position" line) and then you can add `xwalk` to the exec_script whitelist in another one (and explain why it is needed).

rakuco:
And there's nothing Linux-specific here, so I'd retitle the PR and not mention Linux in the commit where you change the exec_script whitelist.

rakuco:
One last thing: don't you also need to port our chromium-crosswalk changes to GN as well? The following commits touch gyp files:
* c72323e87e0b3f72594c8a399b8d6caea1bef167 ("[Extensions] Build glib message pump on Ozone by default")
* 29782db8bf689df501d3714b1f49fa3831bcb1b0 ("Add flag to disable hrtf in webaudio")
* 84686b5f25740746f883105347de1b898c46b66a ("[Windows] Support RealSense Cameras")
* 63e1f4143742cffcdb05438f38b715a914dc9352 ("[Temp] Do not enable generate_multidex_config by default on Android.")
* 2b2cbc8c8344a9a154ac5f477b8e59edda39e9e9 ("[Windows] Implementation of 'AudioDestinationNode.devicePosition' attribute")
* ad2a7f0f00c95fad6fa8c2f8484fd63fb7572fc4 ("[Temp] Make |enable_webvr==0| work.")

Most of those are not related to Linux so you don't necessarily have to port all of them here, but I thought it'd be good to bring them up in case you were not considering them.

heke123:
@rakuco   Thanks very much for remind me this.
for this one:
     2b2cbc8 ("[Windows] Implementation of 'AudioDestinationNode.devicePosition' attribute")
The title is for windows only, but seems the gyp modification has affects on all platform, should I port it now?

   And below 2 commits will be port into GN soon.
   c72323e ("[Extensions] Build glib message pump on Ozone by default")
   29782db ("Add flag to disable hrtf in webaudio")

rakuco:
> The title is for windows only, but seems the gyp modification has affects on all platforms, should I port it now?

Yes, I think you'll need to port this one too (it should be a matter of adding a few lines to some BUILD.gn files).

rakuco:
> 63e1f41 ("[Temp] Do not enable generate_multidex_config by default on Android.")

There doesn't seem to be anything to do with this one: as far as I can see, each target needs to set `enable_multidex = true` for the multidex stuff to be enabled (i.e. it becomes opt-in instead of opt-out, which is what we want)

rakuco:
FYI, I'm handling the Android-specific changes in #377.

heke123:
@rakuco  
2b2cbc8 ("[Windows] Implementation of 'AudioDestinationNode.devicePosition' attribute")
It contains the BUILD.gn.  so we don't need do anything.

You intent to remove below 2 commits,  I guess we don't need to port them either?
c72323e ("[Extensions] Build glib message pump on Ozone by default")
29782db ("Add flag to disable hrtf in webaudio")

rakuco:
> 2b2cbc8 ("[Windows] Implementation of 'AudioDestinationNode.devicePosition' attribute")
> It contains the BUILD.gn. so we don't need do anything.

Excellent!

> You intent to remove below 2 commits, I guess we don't need to port them either?
> c72323e ("[Extensions] Build glib message pump on Ozone by default")
> 29782db ("Add flag to disable hrtf in webaudio")

Yep, correct.

rakuco:
yes! victory at last :+1:

heke123:
Thanks so much for your help!

PR #376, opened 2016-08-24
M53: Fix GYP for Android.

msisov:
It partly works. At least, gyp_xwalk can be executed and almost half of the source code files can be compiled.


mrunalk:
Have you submitted a patch upstream for this? Otherwise, I don't think we can land these patches just to make Android build work on gyp. We will have to wait until GN build is working on Android.

rakuco:
I think it was more of an experiment to see how much the gyp files would need to be changed in case the GN code wasn't ready soon.
@msisov ^

msisov:
Yes, true. At first, I thought I'm able to build the Android build completely, but then it turned out more fixes are needed. I guess we can discard this PR.

PR #377, opened 2016-08-24
[Android] Make it possible to build content_shell_apk with GN.

rakuco:
This pull request contains a few separate patches, all of which are required in order to be able to build the `content_shell_apk` target using GN with Crosswalk's local modifications on top of the upstream commits.

* "[Temp] Make |enable_webvr==false| work in GN" and "[Android] Add ApplicationStatusManager.java to the GN build" just add some new or moved files to the respective `BUILD.gn` files. Their commit messages indicate which commits they should be squashed into in the future.
* "[Backport] GN: Set |enable_webvr| in the declare_args() block" is a backport from a commit I landed upstream that makes it possible to override the value of `enable_webvr` (which we currently need to disable in Crosswalk).

Linux and Windows are out of the scope of this pull request: they will be handled separately in other PRs.

RELATED BUG=XWALK-3674

### DIFF
--- a/base/BUILD.gn
+++ b/base/BUILD.gn
@@ -2187,6 +2187,7 @@ if (is_android) {
       "android/java/src/org/chromium/base/ApiCompatibilityUtils.java",
       "android/java/src/org/chromium/base/ApkAssets.java",
       "android/java/src/org/chromium/base/ApplicationStatus.java",
+      "android/java/src/org/chromium/base/ApplicationStatusManager.java",
       "android/java/src/org/chromium/base/BaseChromiumApplication.java",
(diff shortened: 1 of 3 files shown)


rakuco:
@darktears @lincsoon @heke123: PTAL. While He Ke handles the Linux and common parts of the GN porting work, I'm trying to speed up the porting process for other platforms (especially since we'll likely need GN on Android to be working before we move to M53).

@mrunalk @msisov: FYI. These changes will be needed in M53. The ones which are not backports can be squashed directly into the original commits (it's probably a good idea to edit the original commits' commit messages and add a note saying "M53 rebasing: added GN support" or something like that).

mrunalk:
LGTM

PR #379, opened 2016-08-26
[Backport] Change back to destroy the CDM synchronously

wuhengzhi:
Outstanding promises may get rejected due to blink garbage collection
destroying all the remaining EME objects when they are no longer
referenced. The original fix destroyed the CDM asynchronously.
However, that causes problems as the CDM may hold references to
objects owned by RenderFrameImpl, and they may get accessed after
~RenderFrameImpl.

This fix posts a task to reject the promise asynchronously. That way
any outstanding promises that need to be rejected are handled after
gc is done. It affects all EME promise rejections (from Chromium).

Resolving promises is still done synchronously as there may be events
already posted that need to happen only after the promise is resolved.

BUG=597355
TEST=existing EME tests still pass

Review-Url: https://codereview.chromium.org/1987883002

BUG=XWALK-7221

### DIFF
--- a/media/blink/cdm_session_adapter.cc
+++ b/media/blink/cdm_session_adapter.cc
@@ -26,26 +26,12 @@ namespace {
 const char kMediaEME[] = "Media.EME.";
 const char kDot[] = ".";
 const char kTimeToCreateCdmUMAName[] = "CreateCdmTime";
-
-void ReleaseCdm(scoped_refptr<MediaKeys> cdm) {
-  // |cdm| will be freed now.
-}
-
 }  // namespace
 
 CdmSessionAdapter::CdmSessionAdapter()
     : trace_id_(0), weak_ptr_factory_(this) {}
 
-CdmSessionAdapter::~CdmSessionAdapter() {
-  // Freeing |cdm_| may take a while, so post a task to do it asynchronously.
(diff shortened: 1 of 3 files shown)


wuhengzhi:
@rakuco This PR is to fix the crash issue when exit app, purely cherry-picked from upstream.

rakuco:
Great, thank you. There's just one thing to fix: can you remove the "Cr-Commit-Position" line from your commit message: having it will confuse the script that gets the latest upstream commit for the devtools code since it will point to a commit that does not exist upstream.

wuhengzhi:
Removing, thanks.

rakuco:
Hmm, I've just noticed you're not targeting master here. Since that upstream commit is from M53, don't you need to merge it to master (M52), crosswalk-22 (M52) and crosswalk-21 (M51)?

wuhengzhi:
This issue has been resolved in M52. Please see https://bugs.chromium.org/p/chromium/issues/detail?id=597355#c51

rakuco:
Oh, nice, thanks. Merging, please don't forget to roll `DEPS.xwalk` so that I can manually trigger a new crosswalk-21 build.

wuhengzhi:
Yes, will update it soon.

PR #380, opened 2016-08-26
Remove unused private field device_position_

mrunalk:
../../content/renderer/media/renderer_webaudiodevice_impl.h:94:25: error: private field 'device_position_' is not used
  media::StreamPosition device_position_;
                        ^

### DIFF
--- a/content/renderer/media/renderer_webaudiodevice_impl.h
+++ b/content/renderer/media/renderer_webaudiodevice_impl.h
@@ -91,8 +91,6 @@ class RendererWebAudioDeviceImpl
   // period of silence. We do this on android to save battery consumption.
   base::CancelableClosure start_null_audio_sink_callback_;
 
-  media::StreamPosition device_position_;
-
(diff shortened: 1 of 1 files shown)


mrunalk:
@pozdnyakov PTAL

rakuco:
FWIW, I think this can go directly into master so you get a smaller diff in `next` and also remove an unused parameter from M52 (even if you don't get the warning).

mrunalk:
@rakuco, I have submitted this PR for master now. But wouldn't this create a seprate commit in master which we would have to squash anyway? Feels like additional overhead to me.

rakuco:
It's _some_ overhead, but it's minimal; having it in master means it's test-built and reduces the delta between master and next, which is always a good thing. Given it takes at most a couple of minutes to squash this in next (if you really want to), I think it's an acceptable trade-off.

rakuco:
lgtm, by the way

PR #382, opened 2016-08-31
Merge some GN changes from master to next

msisov:
merge https://github.com/crosswalk-project/chromium-crosswalk/pull/377, https://github.com/crosswalk-project/chromium-crosswalk/pull/378 and https://github.com/crosswalk-project/chromium-crosswalk/pull/381 from master to next


rakuco:
lgtm. One thing I said in the original pull requests is that all commits except for the backports can be squashed into the gyp versions, which you should also consider doing.
And don't forget my comment about backporting an upstream change in #381.

PR #383, opened 2016-09-01
next: Rebase patch controlling the WebVR build

rakuco:
This patch set first reverts "[Temp] Make |enable_webvr==false| work in GN" (see my comment in accdb96d922d789bf8204ff4ca1af87d09673c90) and later implements the same thing in the right files.

You can get rid of both the first commit and accdb96d922d789bf8204ff4ca1af87d09673c90, as only "[Temp] Only depend on deviecs/vr when enable_webvr is on" is required in M53.

### DIFF
--- a/content/public/android/BUILD.gn
+++ b/content/public/android/BUILD.gn
@@ -40,7 +40,6 @@ android_library("content_java") {
     "//device/usb:java",
     "//device/vibration:mojo_bindings_java",
     "//device/vibration/android:vibration_manager_android",
-    "//device/vr:java",
     "//media/base/android:media_java",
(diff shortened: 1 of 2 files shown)


rakuco:
@mrunal @msisov

mrunalk:
@rakuco I have removed the commit so you don't need to revert it now but make sure to rebase your PR.

rakuco:
I'm no longer at work today, so I'll only be able to do that tomorrow. You can also just cherry-pick the patch if you prefer.

rakuco:
rebased, merging.

rakuco:
I'll take care of rolling DEPS.xwalk.

PR #385, opened 2016-09-29
[Backport][Windows] Pass "/MD" and variants outside the "runtime_library" target

rakuco:
This patch is required for the .NET extension bridge to be built.

RELATED BUG=XWALK-7336

> Create separate configurations for passing "/MD", "/MT" and their debug
> variants: "dynamic_crt" passes "/MD" and "/MDd", whereas "static_crt"
> passes "/MT" and "/MTd". BUILDCONFIG then depends on "default_crt",
> which has some logic to choose whether to use either dynamic_crt or
> static_crt.
>
> The main reason behind this is to allow users to config -= the
> "default_crt" config: in gyp, it was possible to configure what was
> going to be passed to the compiler via the
> "win_{release,debug}_RuntimeLibrary" variable, which was useful when
> building code that needs to pass "/CLR", as it requires "/MD" and does
> not work with "/MT".
>
> R=brettw@chromium.org,dpranke@chromium.org,scottmg@chromium.org
>
> Review-Url: https://codereview.chromium.org/2379573003

### DIFF
--- a/build/config/BUILDCONFIG.gn
+++ b/build/config/BUILDCONFIG.gn
@@ -461,6 +461,7 @@ _native_compiler_configs = [
 ]
 if (is_win) {
   _native_compiler_configs += [
+    "//build/config/win:default_crt",
     "//build/config/win:lean_and_mean",
     "//build/config/win:nominmax",
     "//build/config/win:unicode",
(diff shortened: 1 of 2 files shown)


rakuco:
@heke123 @darktears

heke123:
LGTM

PR #386, opened 2016-09-30
[Windows] GN: Add missing RSSDK files in media/.

rakuco:
This commit should be squashed into "[Windows] Support RealSense
Cameras" in the future. It simply does the same thing in GN: a new
argument `use_rssdk` has been added (defaulting to `false`) and takes
care of selecting which RSSDK files to include.

This is an improved version of a patch originally written by Ke He
<ke.he@intel.com>.


rakuco:
@darktears @heke123 
@mrunalk and @msisov: fyi

darktears:
LGTM

PR #391, opened 2016-10-06
crosswalk-22: Fix the build after e762df3

rakuco:
Remove a stray line that should have been removed in e762df3 ("Revert
AudioDestinationNode.devicePosition commits"). It was preventing the
build from succeeding.

RELATED BUG=XWALK-7030
RELATED BUG=XWALK-7226


baleboy:
LGTM

PR #394, opened 2016-10-18
Revert "Replace WebAudio MediaCodec usage with FFmpeg" commit

xzhan96:
This reverts the following commit:
* c0baf4af ("Replace WebAudio MediaCodec usage with FFmpeg. A ~4x improvement.")

This code was reverted because Crosswalk disable the unified media pipeline
in the commit 18c4d465 of xwalk.

BUG=XWALK-7307

### DIFF
--- a/content/browser/renderer_host/render_message_filter.cc
+++ b/content/browser/renderer_host/render_message_filter.cc
@@ -86,6 +86,7 @@
 
 #if defined(OS_ANDROID)
 #include "content/browser/media/android/media_throttler.h"
+#include "media/base/android/webaudio_media_codec_bridge.h"
 #endif
 
 #if defined(OS_MACOSX)
@@ -120,7 +121,14 @@ void NoOpCacheStorageErrorCallback(
     std::unique_ptr<CacheStorageCacheHandle> cache_handle,
     CacheStorageError error) {}
 
-}  // namespace
+}
+
+#if defined(OS_ANDROID)
+void CloseWebAudioFileDescriptor(int fd) {
+  if (close(fd))
+    VLOG(1) << "Couldn't close output webaudio fd: " << strerror(errno);
+}
+#endif  // namespace
 
 RenderMessageFilter::RenderMessageFilter(
     int render_process_id,
@@ -221,6 +229,9 @@ bool RenderMessageFilter::OnMessageReceived(const IPC::Message& message) {
                         OnPreCacheFontCharacters)
 #endif
     IPC_MESSAGE_HANDLER(ViewHostMsg_MediaLogEvents, OnMediaLogEvents)
+#if defined(OS_ANDROID)
+    IPC_MESSAGE_HANDLER(ViewHostMsg_RunWebAudioMediaCodec, OnWebAudioMediaCodec)
+#endif
     IPC_MESSAGE_UNHANDLED(handled = false)
   IPC_END_MESSAGE_MAP()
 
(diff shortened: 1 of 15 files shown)


xzhan96:
@rakuco  please review this patch, thanks

rakuco:
From XWALK-7307 it is not clear if this is the only required fix; this seems related to upstream's switch to ffmpeg, whereas the bug report mentions issues dating back to Crosswalk 19. Can you elaborate?

rakuco:
Plus you mean "Revert" in the title, not "Rever".

xzhan96:
>   no_proxy=ostc.intel.com,localhost,127.0.0.1
 using PTY: True
Usage: gclient.py sync [options]
gclient.py: error: no such option: --manually_grab_svn_rev

@rakuco do you know why the code check always fail?

rakuco:
I do, the slaves need to be restarted to include a fix, hang in there.

xzhan96:
> From XWALK-7307 it is not clear if this is the only required fix; this seems related to upstream's switch to ffmpeg, whereas the bug report mentions issues dating back to Crosswalk 19. Can you elaborate?

There are two ways to decoding web audio, 1) ffmpeg and 2) MediaCodec. As we have disable ffmpeg pipeline, the only way for audio playback is using MediaCodec, this patch  https://bugs.chromium.org/p/chromium/issues/detail?id=424174 disable all the MediaCodec related code for the project spitzer. And it was merged before Chromium 49(Crosswalk 19), which is the reason of the issue appears since 19.

This case is related with XWALK-7231, but not totally the same. The reason why the patch of XWALK-7231 can't solve this issue is that WebAudio use a separate pipeline instead of MediaPlayer, as you can see audio_decoder_android.cc in https://bugs.chromium.org/p/chromium/issues/detail?id=424174.

rakuco:
> I do, the slaves need to be restarted to include a fix, hang in there.

Done, please try again.

rakuco:
> And it was merged before Chromium 49(Crosswalk 19), which is the reason of the issue appears since 19.

I see. Does it mean we'll need to backport this all the way to Crosswalk 21?

xzhan96:
> I see. Does it mean we'll need to backport this all the way to Crosswalk 21?
Sure if we need support this feature.

rakuco:
> Sure if we need support this feature.

What feature? I thought this was about fixing a bug that's present in all releases since 19.

xzhan96:
> What feature? I thought this was about fixing a bug that's present in all releases since 19.

Sorry the misused.
You are right. It's a bug fix

rakuco:
So again: will this need to be backported all the way back to Crosswalk 21, which is the oldest release we currently support?

rakuco:
I'm asking in particular because this is a 1000-line patch, which means it's quite risky to introduce in an older release at this point.

xzhan96:
As we already shipped stable version from 19 to 21, I think its proper only backport to Crosswalk 22, customer can use 22 or later to avoid the audio issue, do you think so?

rakuco:
We can discuss that with @baleboy in the bug. For now, let's get it into 23 and 22 at least. Don't forget you need to roll `DEPS.xwalk`.

xzhan96:
code check issue:
The media directory requires source formatting. Please run git cl format media

BernsteinA:
Would this reopen https://groups.google.com/a/chromium.org/forum/m/#!topic/chromium-bugs/i0ZiRaJWpNU for crosswalk users?

baleboy:
@BernsteinA yes it looks like that

BernsteinA:
@baleboy That's disappointing. Is this the best place to discuss that decision?

baleboy:
@BernsteinA The related Jira issue XWALK-7307 is a better place for that. Right now the choice is between not having audio decode at all or having it but slower.

BernsteinA:
That's not quite accurate. Will comment on Jira when I get to work

PR #399, opened 2016-11-14
Stop calling Free on Slot

msisov:
I've touched base/threading/thread_local.h and content/browser/android/content_view_core_impl.cc.
Then did a merge to c5cf474 and  	f13a5b7. That is why, I guess, it shows many commits here.

### DIFF
--- a/base/threading/thread_local.h
+++ b/base/threading/thread_local.h
@@ -116,7 +116,7 @@ template <typename Type>
   public:
    ThreadLocalPointer() {}
 
-   ~ThreadLocalPointer() { slot_.Free(); }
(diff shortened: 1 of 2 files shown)


rakuco:
It looks like you need to rebase your checkout.

mrunalk:
@msisov You should keep your changes separate from the original commits until we do final merge into master so that we know what's changing. You can reference the original commits in your commit messages. Also if they are modifying separate original commits then you should have respective number of commits so it will be easier to merge them later.

msisov:
check now, please

rakuco (approved):
lgtm, thanks!

PR #401, opened 2016-12-06
BUG=XWALK-7004

dlmu2001:
Support Crosswalk and Android System WebView in the same application.
Remove check and load icudtl.dat from assets folder firstly,since
xwalk put icudtl.dat in other folders

### DIFF
--- a/content/browser/renderer_host/render_message_filter.cc
+++ b/content/browser/renderer_host/render_message_filter.cc
@@ -86,6 +86,7 @@
 
 #if defined(OS_ANDROID)
 #include "content/browser/media/android/media_throttler.h"
+#include "media/base/android/webaudio_media_codec_bridge.h"
 #endif
 
 #if defined(OS_MACOSX)
@@ -120,7 +121,14 @@ void NoOpCacheStorageErrorCallback(
     std::unique_ptr<CacheStorageCacheHandle> cache_handle,
     CacheStorageError error) {}
 
-}  // namespace
+}
+
+#if defined(OS_ANDROID)
+void CloseWebAudioFileDescriptor(int fd) {
+  if (close(fd))
+    VLOG(1) << "Couldn't close output webaudio fd: " << strerror(errno);
+}
+#endif  // namespace
 
 RenderMessageFilter::RenderMessageFilter(
     int render_process_id,
@@ -221,6 +229,9 @@ bool RenderMessageFilter::OnMessageReceived(const IPC::Message& message) {
                         OnPreCacheFontCharacters)
 #endif
     IPC_MESSAGE_HANDLER(ViewHostMsg_MediaLogEvents, OnMediaLogEvents)
+#if defined(OS_ANDROID)
+    IPC_MESSAGE_HANDLER(ViewHostMsg_RunWebAudioMediaCodec, OnWebAudioMediaCodec)
+#endif
     IPC_MESSAGE_UNHANDLED(handled = false)
   IPC_END_MESSAGE_MAP()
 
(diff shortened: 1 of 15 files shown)


baleboy:
@lincsoon @fujunwei @rakuco

base/i18n/icu_util.cc
```diff
@@ -86,15 +86,19 @@ void LazyInitIcuDataFile() {
...
-  int fd = base::android::OpenApkAsset(kAndroidAssetsIcuDataFileName,
-                                       &g_icudtl_region);
-  g_icudtl_pf = fd;
-  if (fd != -1) {
-    return;
-  }
-// For unit tests, data file is located on disk, so try there as a fallback.
-#endif  // defined(OS_ANDROID)
+//It's not correct to check and read from assets folder first, since xwalk
+//put icudtl.dat in other folders like res/raw instead of assets.Reading
```
fujunwei: Should we move icudtl.dat to assets directory align with XWalkRuntimeLib.apk that hold resources in assets directory.

dlmu2001: If icudtl.dat was moved to assets, it'll have coexistence issue with stock android Webview. Since there's only one assets folder for application, once app call android webview's function, it'll copy android webview's icudtl.dat to assets.  The two icudtl.dat will conflict.

dlmu2001: And comments' in ResourceExtractor.java

"// Loading "icudtl.dat" from "assets/"" currently does not work with either
// embedded mode (the file is in raw/res) or shared mode (the app's context is
// used to retrieve the AssetManager, not Crosswalk's). We thus need to put
// those special files in a different directory so that we leverage the fallback
// code in Chromium to load these files from disk

fujunwei: How about using `#if 0` to disable these codes?

fujunwei: Refine descriptions of commit with https://www.chromium.org/developers/contributing-code#TOC-Writing-change-list-descriptions

dlmu2001: Should the new commit include in the same PR?

fujunwei: You can use the option of `--amend` to refine the commit.
`git commit -a --amend`

dlmu2001:
Last time I use --amend, but it seems can't do merge well and trybot failed.

2016-12-09 10:59 GMT+08:00 fujunwei <notifications@github.com>:

base/i18n/icu_util.cc
dlmu2001: Last time I use --amend,  it seems merge can't do  well and trybot failed.

base/i18n/icu_util.cc
```diff
@@ -86,6 +86,11 @@ void LazyInitIcuDataFile() {
   if (g_icudtl_pf != kInvalidPlatformFile) {
     return;
   }
+//It's not correct to check and read from assets folder first, since xwalk
+//put icudtl.dat in other folders like res/raw instead of assets.Reading
+//from assets has a risk of conflict with android system webview.
+//See https://crosswalk-project.org/jira/browse/XWALK-7004
+#if 0
 #if defined(OS_ANDROID)
```
fujunwei, line 89: Only keep `#if 0`
`#if defined(OS_ANDROID)` => `#if 0`

dlmu2001: Is it code style requirement?
I don't think it's a good practice to replace #if(OS_ANDROID） with #if 0.
Why not remove them in such case?

dlmu2001: By the way, it seems the trybot will fail in merge check once I use git commit --amend.

fujunwei: Please try to combine them into one commit.
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

fujunwei:
https://github.com/crosswalk-project/chromium-crosswalk/pull/401/commits/625ff15da80d8da35645fbff6f715b80cedc521f you need to remove it.

dlmu2001:
Once I re-based and execute "git push" ,it always ask me to merge. So there's 625ff15.

dlmu2001:
Can I use git reset --hard to remove it?

fujunwei:
You can create a new local branch, and then cherry-pick it
git checkout -b local_branch origin/master
git cherry-pick commit_id
git push -f dlmu2001 local_branch:master

dlmu2001:
I did not make branch in remote before. So the cherry-pick mean seems not work.
I make a branch and delete the commit . 
Sorry for it trigger the new PR.

PR #402, opened 2016-12-09
Support Crosswalk and Android System WebView in the same application

dlmu2001:
xwalk will put icudtl.dat in different folder like raw/res instead of
assets.So it's not right to check and read from assets/icudtl.dat in
high priority.Otherwise when android webview is called , app will
extract webview.apk and copy it's icudtl.dat to assets,hence the two
icudtl.dat will conflict.

BUG=XWALK-7004

### DIFF
--- a/base/i18n/icu_util.cc
+++ b/base/i18n/icu_util.cc
@@ -86,6 +86,11 @@ void LazyInitIcuDataFile() {
   if (g_icudtl_pf != kInvalidPlatformFile) {
     return;
   }
+//It's not correct to check and read from assets folder first, since xwalk
+//put icudtl.dat in other folders like res/raw instead of assets.Reading
+//from assets has a risk of conflict with android system webview.
+//See https://crosswalk-project.org/jira/browse/XWALK-7004
+#if 0
 #if defined(OS_ANDROID)
   int fd = base::android::OpenApkAsset(kAndroidAssetsIcuDataFileName,
                                        &g_icudtl_region);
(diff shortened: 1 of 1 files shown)


fujunwei:
LGTM

dlmu2001:
@lincsoon @baleboy  @rakuco

New thread here, 401 has been closed.

baleboy:
@lincsoon is it OK to merge this? @fujunwei gave his OK

rakuco:
He's now @sunlin-link FWIW

baleboy:
@rakuco thanks, actually I should have asked you since this is chromium-crosswalk

rakuco:
I think this needs some more work, as well as review from the Android people.

For one, I'm not sure if we've ever tested or explicitly supported apps mixing the stock WebView with Crosswalk. Second, we also need to make sure this doesn't break apps using only Crosswalk.

With that in mind, I wonder if there's anything to fix on the Crosswalk side to avoid changing chromium-crosswalk, and, lastly, just like Chromium we don't leave dead, commented out code in the tree, so if that's the path we want to go that code needs to be removed, not commented out.

@sunlin-link: ping again

dlmu2001:
@rakuco 

Xwalk  put icudtl.dat in different folder like raw/res instead of assets now. It's not reasonable to check and read from assets folder. It's obvious and will not break apps using only Crosswalk.

By the way, the process is so long now. I guess this problem will hang for long time in your suggestion.

zhanghai:
It's March 9th now and please take a merge - it's impeding our release, considering that we cannot use `XWalkView` in an Android `Service` so we have to use `WebView`.

